PR #295, opened 2025-02-27
add optional raise error when fail to run_analysis

kaihsin:


src/kirin/analysis/forward.py
```diff
@@ -35,6 +35,7 @@ def run_analysis(
         self,
         method: ir.Method,
         args: tuple[LatticeElemType, ...] | None = None,
+        raise_error_if_failed: bool = False,
```
Roger-luo, line 38:
```suggestion
        no_raise: bool = True,
```

I think it's usually called this?

Roger-luo (changes requested):

PR #301, opened 2025-02-28
Fix bug on Call and GetField

kaihsin:
This PR address #300

1. Change GetField from Pure() to MaybePure() because a field  is specify  by integer which is attribute not argument, and the return value might be Unknown(), and there is no dataflow for this stmt. 
2. Fix constprop interp method for GetField to also check if the method fields are lattice element or not.


Roger-luo:
See the fix in #302 I think that's the right fix instead of changing the purity model of `getfield`

PR #303, opened 2025-03-01
run black

Roger-luo:
not sure why black check was not passing


JerryChen97:
Until Lint#651, your CI has been using black 24.10. After that its been changed to 25.1 probably why it suddenly failed with multiple files.

PR #311, opened 2025-03-10
Add trait to Attribute base class. 

weinbe58:
this PR adds a new class variable `traits` to the attribute base class. Allowing one to annotate attribute with trait. This is a change we use for supporting the text format.


Roger-luo (approved):

Roger-luo:
just a side note, @weinbe58  we should always use `<name>/<branch>` for PRs for the sake of easier looking up for different developer's branch.

PR #312, opened 2025-03-10
Registry for `PyClass` values. 

weinbe58:
This registry is required for things like Parsing so that there is some way of mapping the string representation of a `PyClass` object to a Type binding. For example when constructing a Function signature from the string: `(type1, type2, ...) -> type3` I need to know how to construct the type binding from the string representation. This PR also allows ones to customize how `PyClass` is printed to avoid naming collisions between different classes, this can be done either by a new prefix or a custom display name for the type. 

A side effect is that going forward it would be good to construct the Type binding using this method directly. For example

```python
from ._dialect import dialect

class Qubit:
   pass

QubitType = dialect.register_py_class(Qubit, ...)

```

### DIFF
--- a/test/ir/test_dialect.py
+++ b/test/ir/test_dialect.py
@@ -1,0 +1,34 @@
+import pytest
+
+from kirin.ir import Dialect
+from kirin.types import PyClass
+
+
+def test_py_type_register():
+
+    class TestClass:
+        pass
+
+    dialect = Dialect("test")
+
+    TestType = dialect.register_py_type(
+        TestClass, display_name="TestType", prefix="test"
+    )
+    assert TestType == PyClass(TestClass, prefix="test", display_name="TestType")
+
+    assert dialect.python_types == {("test", "TestType"): TestType}
+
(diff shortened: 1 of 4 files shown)


Roger-luo (changes requested):
need a simple test somewhere.

test/ir/test_dialect.py
```diff
@@ -0,0 +1,37 @@
...
+        dialect.register_py_type(TestClass, display_name="TestClass", prefix="test")
+
+    with pytest.raises(ValueError):
+        dialect.register_py_type(
+            TestClass, display_name="TestType", prefix="other_prefix"
+        )
+
+
+if __name__ == "__main__":
+    test_py_type_register()
```
Roger-luo, line 37:
```suggestion
```

Roger-luo (approved):

Roger-luo:
@weinbe58 need to fix the linter otherwise LGTM

Roger-luo:
@weinbe58 might due to an out dated version of black you installed locally?

PR #315, opened 2025-03-11
Split LoweringState

weinbe58:
Split up SSA-related APIs of `LoweringState` to a mix-in class.


Roger-luo (approved):

Roger-luo:
So after thinking about this for a bit more time I think we should cleanup the lowering before merging the text format parser in. Otherwise there could be a bigger change in the text parser. 

And it would be helpful if we can have a PR somewhere showing how @weinbe58 intends to be using the lowering state somewhere then we can come up with a cleaner split.

PR #316, opened 2025-03-12
Inline `len` for `IList`

weinbe58:
Some kinds of `IList` can't be folded but the length can be determined at compile time. This PR adds a rewrite that inlines the `len` if the length can be determined by type inference. 

This is preventing some kinds of bloqade programs from being folded to QASM2. The list of qubits can't be folded to a constant value because there is no concrete interpreter for `QRegGet`. 

```python


@qasm2.extended
def _log_ghz(i_layer: int, qubits: ilist.IList[qasm2.Qubit, Any]):
    stride = len(qubits) // (2**i_layer)

    if stride == 0:
        return
    
    offset = stride // 2
    for j in ilist.range(0, len(qubits), stride):
        qasm2.cx(ctrl=qubits[j], qarg=qubits[j + offset])

    _log_ghz(i_layer + 1, qubits)
    

@qasm2.extended
def main():
    q = qasm2.qreg(6)
    _log_ghz(0,[q[0], q[2], q[4], q[1], q[3], q[5]])
    

fold = QASM2Fold(qasm2.extended)
fold(main)
main.print()
fold(main)
```

The second fold fails because of this.


src/kirin/dialects/ilist/rewrite/unroll.py
```diff
@@ -129,3 +130,10 @@ def rewrite_ForEach(self, node: ForEach) -> RewriteResult:
 
         node.delete()
         return RewriteResult(has_done_something=True)
+
+    def rewrite_Len(self, node: Len):
```
Roger-luo, line 134: why does this needs to live inside unroll rule? it has nothing to do about ilist unroll?

weinbe58: I can move it but I will steal that function to get the length of from the type attribute

weinbe58: Arguably though this is required for loop unrolling so it kind of makes sense

weinbe58: I've move it to `py.len` dialect.

src/kirin/dialects/py/len.py
```diff
@@ -55,3 +57,50 @@ def lower_Call_len(
...
+
+class InferLen(RewriteRule):
+    EllipsisType = types.PyClass(type(...))
+
+    @cached_property
+    def Constant(self):
+        # Avoid circular import caching the result
+        from kirin.dialects.py.constant import Constant
+
+        return Constant
```
Roger-luo, line 70: what is this for?

src/kirin/dialects/py/len.py
```diff
@@ -55,3 +57,50 @@ def lower_Call_len(
         return lowering.Result(
             state.append_stmt(Len(value=state.visit(node.args[0]).expect_one()))
         )
+
+
+class InferLen(RewriteRule):
+    EllipsisType = types.PyClass(type(...))
```
Roger-luo, line 63: why do you need Ellipsis?

src/kirin/dialects/py/len.py
```diff
@@ -55,3 +57,50 @@ def lower_Call_len(
...
+        if (
+            coll_type.is_subseteq(self.IListType)
+            and isinstance(coll_type.vars[1], types.Literal)
+            and isinstance(coll_type.vars[1].data, int)
+        ):
+            return coll_type.vars[1].data
+        elif coll_type.is_subseteq(types.Tuple) and not any(
+            var.is_subseteq(self.EllipsisType) for var in coll_type.vars
+        ):
+            return len(coll_type.vars)
```
Roger-luo, line 94: this is wrong, why do we need to deal with `tuple` at all inside `ilist` dialect? Isn't this case being covered by constprop?

Roger-luo: and you are mixing up Python types with Kirin types here... You want `types.Vararg` not `...`

Roger-luo (changes requested):
why don't we just put a hint into the `SSAValue` with a `IList[XXX, Literal(something)]` type then run constant fold again (which only rewrites those value with constant hint instead of running the entire constant prop analysis)

src/kirin/dialects/py/len.py
weinbe58: I want to make sure none of the types inside the tuple isn't `Ellipses` , e.g., `Tuple[int, ...]` because that doesn't have a well defined length even though it has two parameters.

weinbe58: > this is wrong, why do we need to deal with `tuple` at all inside `ilist` dialect? Isn't this case being covered by constprop?

I'm handing these cases separately. I'm not handling nested types. Constant prop doesn't always run because of missing concrete interpreters but we can get this from type inference.

Roger-luo: Constant prop uses a PartialTuple lattice element to keep track of tuples. This includes the length of a tuple despite its value.

Roger-luo: You are dealing with Kirin's type system now so you don't have ellipsis there's only Vararg type in generics.

Roger-luo:
Pass or analysis. Remember you cannot modify code in analysis because that goes through functions you don't own.

src/kirin/rewrite/infer_len.py
```diff
@@ -0,0 +1,41 @@
...
+from kirin.rewrite.abc import RewriteRule, RewriteResult
+
+
+class InferLen(RewriteRule):
+
+    @cached_property
+    def IListType(self):
+        from kirin.dialects import ilist
+
+        return ilist.IListType
```
Roger-luo, line 15: I don't understand what is this for? Also shouldn't this rewrite pass be inside `ilist`? It does not necessarily always happen during an aggressive fold.

Roger-luo (commented):
I think I'll do some rework on this pass in this branch a bit later before merging.

src/kirin/rewrite/infer_len.py
weinbe58: The problem is circular imports and it seems bad to have to import this type every time a method is called.

weinbe58: If it is not needed for const folding then we don't need this rewrite. 

The only case I can think of if you are aggressively folding a kernel with an argument that has a type annotation that exactly bounds the size of that input. In that case you can unroll the kernel if you're loop bounds depend on the size of the argument not necessarily the argument itself. 

it is a pretty small edge case so sure we can move it into ilist to prevent the circular import problems.

Roger-luo:
for a side note, I'm thinking about adding canonicalize + inference hook to each dialect where `canonicalize` is a list of rewrites from each dialect to run with `Canonicalize` pass and `inference` are list of passes to run within `typeinfer` pass after this PR to clean things up.

PR #318, opened 2025-03-13
`StmtTrait -> Trait[T]` + updating attribute Traits

weinbe58:
Per discussion with @Roger-luo this PR renames the base class of Traits and also makes it generic for better hinting. I also realized my last PR on Attribute traits was missing an implementation of `get_trait` so this PR also adds that as well.

### DIFF
--- a/src/kirin/ir/attrs/abc.py
+++ b/src/kirin/ir/attrs/abc.py
@@ -50,6 +50,21 @@ class Attribute(ABC, Printable, metaclass=AttributeMeta):
     @abstractmethod
     def __hash__(self) -> int: ...
 
+    @classmethod
+    def has_trait(cls, trait_type: type[Trait["Attribute"]]) -> bool:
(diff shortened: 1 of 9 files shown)


src/kirin/ir/attrs/abc.py
```diff
@@ -41,10 +42,27 @@ class Attribute(ABC, Printable, metaclass=AttributeMeta):
...
         default=frozenset(), init=False, repr=False
     )
     """Set of Attribute traits."""
 
     @abstractmethod
     def __hash__(self) -> int: ...
+
+    TraitType = TypeVar("TraitType", bound=Trait["Attribute"])
+
+    def get_trait(self, trait: type[TraitType]) -> Optional[TraitType]:
```
Roger-luo, line 55: Can we also have am has_trait method?

Roger-luo (approved):
Just need to add an has_trait method then this is good to go!

Issue #199, opened 2025-02-06
lark bnf spec emitter (a.k.a the text format)


Roger-luo:
Sometimes one wants to construct an exact form of the IR instead of using the Python decorator frontend. It can be quite annoying constructing the IR via constructors, so text format is more favorable in this case. 

This can be done by implementing a code generator that generates the lark spec file (so one can change the syntax of a specific statement like MLIR) and then let lark parse the text format (we are not aiming to parse large code base here anyways) then lower to the IR objects in Python.

weinbe58:
Support for some kinds of statement attributes will be tricky, e.g. simple built-in types should be easy to support but more Python objects will take some thought.

Roger-luo:
this can be done similar to how codegen works - the attribute can contain a custom method implementing how lark files will be generated vive versa.

Roger-luo:
sketch from today's meeting with @weinbe58 

```python
class LarkFile:
    grammar_registry: dict[ir.Statement, str]

class LarkParser:
    dialect: ir.DialectGroup
    grammar_: str
    registry_cons: dict[dialect_str, dict[rule_str, Callable[[lowering.State, LarkTree], ir.Statement]]]
    state: lowering.State


@dialect.register
class FromLarkAST:
    
    def from_const(self, tree: LarkTree):
        return ir.Constant(tree.value)


# ^1():
#     %1 = py.Constant(1)
#     %2 = py.Constant(2)
#     %3 = py.Add(%1, %2)
#          cond_br %cond ^2(%3) ^3(%3)
#
# ^3(%4):
#     %5 = py.Constant(3)

# ^2(%3):
#     %4 = scf.if %3 {
#         %5 = py.Constant(3)
#              scf.yield %5
#     } else {
#         %6 = py.Constant(4)
#              scf.yield %6
#     }

dict[str, dict[str, type[Statement]]]

@dataclass
class Statemetn(ASTNode):
    dialect: str
    stmt: str
    args: list[ASTNode]
    regions: list[ASTNode]
    successors: list[ASTNode]

@dataclass
class SSAValue:
    name: str

SSAValue("1")

```

weinbe58:
@Roger-luo If I where to make an AST for Kirin-IR probably should refactor the printer at some point, should I base the AST off of the current printer?

Roger-luo:
yes, I think actually you could also convert directly from LarkTree to the statement without going through the statement. You can have the lark tree as the IR node anyways.

[...]

PR #319, opened 2025-03-13
Lark text parser

weinbe58:
This PR is a PR to address #199 The overall idea is to use a trait to generate the grammer for statements and leave the overall parsing rules for `Region` and `Block` hardcoded into the grammer along with some other rules regarding things like ssa identifiers block identifiers comments, etc. The grammer is collected from a dialect group merging all the rules together coming from the traits of individual statements and attributes inside the dialect group and compiled into a Lark Grammer. The rules form a registry, making an individual rule to an IR node. 

The final grammar is then used to parse a string given an entry point statement: func. Function, Module, etc. After that, the Lark tree will be visited and lowered using the Lowering framework to generate the IR. The registry of lark rules is used by the lowering infrastructure to dispatch the lark tree labels to the specific statements and attributes.

### DIFF
--- a/src/kirin/dialects/lowering/cf.py
+++ b/src/kirin/dialects/lowering/cf.py
@@ -34,7 +34,7 @@ class CfLowering(FromPythonAST):
         none_stmt = frame.append_stmt(py.Constant(None))
 
         body_frame = state.push_frame(
-            Frame.from_stmts(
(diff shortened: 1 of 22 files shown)


weinbe58:
gonna close this since it is quite old

PR #332, opened 2025-03-24
Review Food Language

johnzl-777:
Doing some homework @Roger-luo assigned me and had some suggestions/fixes for the code examples. 

Suggestions are primarily around grammar/wording, fixes involve things like missing imports/broken code.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -59,17 +59,17 @@ While at the moment only us at [QuEra Computing Inc](https://quera.com) are acti
 
 ### Quantum Computing
 
-Kirin has been used for building several eDSLs within [QuEraComputing](https://quera.com), including:
+Kirin has been used for building several eDSLs within [QuEra Computing](https://quera.com), including:
 
-- [bloqade.qasm2](https://github.com/QuEraComputing/bloqade/tree/main/src/bloqade/qasm2) This is an eDSL for quantum computing that we uses Kirin to define an eDSL for the Quantum Assembly Language (QASM) 2.0. It demonstrates how to create multiple dialects, run custom analysis and rewrites, and generate code from the dialects (back to QASM 2.0 in this case).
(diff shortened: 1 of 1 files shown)


johnzl-777:
I'll start by saying that I did not make it through the entire series about the food language (mostly due to confusion and some frustration) but have opted to allow review of what suggestions I have made. 

Admittedly, I find the writing and the structure to be so jarring/lacking good flow I can't follow it well myself. It feels like there was no attempt to try and go through the tutorials *at least once*. I suppose I am that person this time around but I just find it to be a bit frustrating, especially as I am trying to ramp up my own knowledge of Kirin so I can work on features myself. 

Here are some issues I've encountered (and tried to fix, or will have to fix later). I hope in the future when we create such tutorials we can be a bit more careful/thoughtful:

- Typos - I think this could be mitigated in the future with the usage of a VSCode plugin like Code Spell Checker which has caught many of my own typos very early on.
- Missing Imports - I tried to follow the code examples but inevitably found I had to dig up what the proper imports were myself because they were not given anywhere OR, as in one instance, an import was provided prematurely with no explanation given (I believe `Forward` got imported somewhere and there was no justification for doing it at that particular code point). 
- Unexplained/poorly explained code choices (contributes to the "flow" probably I mention previously)
  - For example, in the very first example of the Food Dialect we have `return x + 1` along with the explanation that we are "doing some math to get a result". I think there should at least be a sentence or two here explaining why that needs to happen, ESPECIALLY when subsequent food examples do NOT have this return arithmetic going on. 
  - With the Control Flow rewrite it would be really useful to see the before and after effect of having the pass versus not having the pass. A simple `main.print()` with and without the rewrite would be perfect for this.
    - `Fixpoint` gets pulled in here as opposed to just using `Walk` for the NewFoodAndNap rewrite, why is that? Is it because we have to worry about nested control flow? If so it should be made explicit
  - In the Food Cost analysis example, all of a sudden `AtLeastXItem(Item)` gets introduced without justification/explanation. I know it will be used in the codegen tutorial but then I feel like we should put the introduction of that in the codegen tutorial and not in the analysis tutorial. 
  - It would be nice to see `Frame` explained in a couple sentences as well, especially with how frequently we have to use `frame.get(...)`

Roger-luo (approved):

Issue #321, opened 2025-03-14
Exception during type inference.


weinbe58:
A problem when running type inference in the following:

```python
from bloqade import qasm2
from kirin.dialects import ilist
from bloqade.qasm2.passes import  QASM2Fold


@qasm2.extended
def log_ghz(qubits):
    n_qubits = len(qubits)
    def _log_ghz(i_layer, qubits):
        stride = n_qubits // (2**i_layer)

        if stride == 0:
            return
        
        offset = stride // 2
        for j in ilist.range(0, n_qubits, stride):
            qasm2.cx(ctrl=qubits[j], qarg=qubits[j + offset])

        _log_ghz(i_layer + 1, qubits)
        
    _log_ghz(0, qubits)
        

@qasm2.extended
def main():
    q = qasm2.qreg(6)
    log_ghz([q[1], q[3], q[5]])
    log_ghz([q[0], q[2], q[4]])
```

gives the following stack trace:

```
Traceback (most recent call last):
  File "/Users/phillipweinberg/Documents/Work/quera/atom_arch/kirin-workspace/test.py", line 7, in <module>
    @qasm2.extended
     ^^^^^^^^^^^^^^
  File "/Users/phillipweinberg/Documents/Work/quera/atom_arch/kirin-workspace/kirin/src/kirin/ir/group.py", line 221, in __call__
    return wrapper(py_func)
           ^^^^^^^^^^^^^^^^
  File "/Users/phillipweinberg/Documents/Work/quera/atom_arch/kirin-workspace/kirin/src/kirin/ir/group.py", line 217, in wrapper
    self.run_pass(mt, *args, **options)
  File "/Users/phillipweinberg/Documents/Work/quera/atom_arch/kirin-workspace/bloqade/src/bloqade/qasm2/groups.py", line 112, in run_pass
    typeinfer_pass(mt)
  File "/Users/phillipweinberg/Documents/Work/quera/atom_arch/kirin-workspace/kirin/src/kirin/passes/abc.py", line 29, in __call__
    result = self.unsafe_run(mt)
             ^^^^^^^^^^^^^^^^^^^
  File "/Users/phillipweinberg/Documents/Work/quera/atom_arch/kirin-workspace/kirin/src/kirin/passes/typeinfer.py", line 19, in unsafe_run
    frame, return_type = self.infer.run_analysis(mt, mt.arg_types)
[...]

PR #333, opened 2025-03-24
fix inference error when input type is a fully parameteric method type

Roger-luo:
partially address #321


Roger-luo:
superceded by #356

PR #342, opened 2025-03-31
feat: add debug dialect

Roger-luo:
this PR adds a new dialect named `debug` to model possible debug related operations, right now this dialect only has an `info` statement that prints a message and optionally some variables feed after the message. Similar to `icecream` we will detect the name of the variable and print the name and value pair of the variable. See below for the printing

image

cc: @cduck

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -187,3 +187,4 @@ main.py
 .ruff_cache
 .python-version
 !package.json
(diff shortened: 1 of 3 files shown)


Roger-luo:
this feature can be backported since it is non-breaking (since we are at 0.x.y this is not violating semver)

PR #358, opened 2025-04-07
fixing bug with multi-return values in statement

weinbe58:
This is a bug that popped up when @johnzl-777 was trying to implement the address analysis on the squin wire dialect. The problem is that before this PR the constprop pass would seemingly fail for no reason due to the default implementation of constprop on statements that return multiple results. This PR addresses this issue in general.


Roger-luo (approved):
Ah good catch we didn't have tests for multi-return values I guess.

weinbe58:
There is no python lowering for this so it makes sense

Roger-luo:
The python lowering is only fixed recently in 0.16 😅

PR #365, opened 2025-04-11
Add Pauli algebra dialect example to docs

david-pl:
There's still an issue with a reference to `Fold`, that mkdocs can't resolve for some reason.


example/pauli/dialect.py
```diff
@@ -0,0 +1,3 @@
+from kirin import ir
+
+dialect_ = ir.Dialect("pauli")
```
Roger-luo, line 3: we usually do the reverse, name the file `_dialect.py` (because you are not interested in using this module outside the package)

name the dialect `dialect` so that you can directly use the module object as if it is the dialect (the APIs in Kirin is smart enough to pick it up)

example/pauli/stmts.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+@statement(dialect=dialect_)
+class Z(PauliOperator):
+    pass
+
+
+@statement(dialect=dialect_)
+class Id(PauliOperator):
+    pass
```
Roger-luo, line 35: not sure if you want to also have it, but fyi. if you want to get rid of those linter error you can implement some dummy `lowering.wraps` to trick Python linter to believe your output is not an SSA but a `np.matrix`

docs/cookbook/paulilang/pauli.md
```diff
@@ -0,0 +1,683 @@
...
+
+
+@statement(dialect=dialect)
+class Id(PauliOperator):
+    pass
+```
+
+
+1. Define the actual dialect to which we want to register our statements.
+2. Notice how we are not registering the `PauliOperator`. That's because we don't actually want users to create an instant of it since it serves just as an abstract type. This is similar to how a user cannot create a number of abstract type `Number`, it's always a type such as `float` or `int`.
```
Roger-luo, line 69:
```suggestion
2. Notice how we are not registering the `PauliOperator`. That's because we don't actually want users to create an instant of it since it serves just as a base type. This is similar to how a user cannot create a number of base type `Number`, it's always a type such as `float` or `int`.
```

we should prob still give pythonistas some respect of wording :-)

docs/cookbook/paulilang/pauli.md
```diff
@@ -0,0 +1,683 @@
...
+3. All Pauli matrices will be pure since they can only return an object of one type. They are also created by a simple call.
```
Roger-luo, line 70: I feel this sentence is not very accurate here:

1. if we think the value of a Pauli operator is `np.matrix` then the output of `pauli.X()` is not always the same because of memory allocation behaviour and limited precision of X. So it is not pure.
2. if we think the value of a Pauli operator is **an unique** `np.matrix`  representing Pauli X operator (a.k.a singletons) then it is pure. Because `pauli.X()` always returns the same value.

Purity is about whether your output is deterministicly only depends on your input afterall, thus there is a subtle difference here. I believe your intend is the 2nd case.

Roger-luo (approved):
just some minor comments I think this is a great addition to the examples!

docs/cookbook/paulilang/pauli.md
david-pl: @Roger-luo I'd agree that we should have the second case here. However, since I'm pulling in a pre-factor into the definitions, we don't always get the same matrix, but rather the same matrix, but multiplied by a number.

So, technically, I'm not really implementing Pauli matrices here, but rather a product of one with a number, which means they're not singletons. Are they still pure then?

I know that's not a very clean implementation, but this way I can implement rewrites on binary multiplications only, which makes the example much easier.

docs/cookbook/paulilang/pauli.md
david-pl: My bad :D

example/pauli/stmts.py
david-pl: I know, but I didn't want to bother with the wraps for the example.

PR #370, opened 2025-04-16
Add get_present_trait method to statement class

david-pl:
Noticed in https://github.com/QuEraComputing/bloqade-circuit/pull/163 that this is useful for linting, but also just for shortening code here and there. I also replaced code that raises a `ValueError` if the result from `get_trait` is `None` here (2nd commit).


Roger-luo (approved):
thanks! Yeah this is a good change.

PR #374, opened 2025-04-19
support stacktrace in interpreter

Roger-luo:
```python

@basic(typeinfer=True)
def some_code_will_error(x):
    """
    This function will raise an error.
    """
    return 1 / x

@basic(typeinfer=True)
def some_func():
    return some_code_will_error(0)

some_func()
```

image


Roger-luo:
this is breaking due to change of error handling. I tried to backport but there are a lot conflict that cannot be backported.

Issue #389, opened 2025-05-01
[Incorrect interpretation] For-loop only runs the first time a function is called, not subsequent calls


cduck:
The first time you run function containing a for-loop over an iterator (here `range(4)`) you get the correct execution but future times, the iterator acts as if it's already exhausted.  I would guess this is a constant propagation or interpreter bug where it it caching mutable runtime objects.

Note: the bug does not occur if optimization is disabled via the decorator `@kirin.prelude.basic_no_opt` or if the iterable value is non-pure or dependent on an argument value.

Tested on:
- kirin-toolchain==0.14.3
- kirin-toolchain==0.17.4

```python
import kirin.prelude

@kirin.prelude.basic
def loop(a: int):
    l = []
    for i in range(4):
    # for i in [1,1,2,3]:  # Same result with this line instead
        l = l + [a]
    return l

print(loop('a'))
print(loop('b'))
print(loop('c'))
```
Output:
```
IList(['a', 'a', 'a', 'a'])
IList()  # WRONG, should be ['b', 'b', 'b', 'b']
IList()  # WRONG, should be ['c', 'c', 'c', 'c']
```

-----

A more complex behavior exposing more interesting bug behavior:
```python
import kirin.prelude

@kirin.prelude.basic
def loop(a: int):
    l = []
    for i in range(1):
        l = l + [a, '1']
    for i in range(2):
        l = l + [a, '2']
    return l

print(loop('a'))
print(loop('b'))
print(loop('c'))
```
```
IList(['a', '1', 'a', '2', 'a', '2'])
IList(['b', '2', 'b', '2'])  # Wrong, should be ['b', '1', 'b', '2', 'b', '2']
IList(['c', '2', 'c', '2'])  # Wrong, should be ['c', '1', 'c', '2', 'c', '2']
```

-----

The bug has the same behavior when `loop` is called multiple times from another kernel:
```python
import kirin.prelude

@kirin.prelude.basic
def loop(a: int):
    l = []
    for i in range(1):
        l = l + [a, '1']
    for i in range(2):
        l = l + [a, '2']
    return l

@kirin.prelude.basic
def main():
    return loop('a') + loop('b') + loop('c')

main()
```
```
IList(['a', '1', 'a', '2', 'a', '2', 'b', '2', 'b', '2', 'c', '2', 'c', '2'])  # WRONG
```

Issue #188, opened 2025-02-02
CF loop unroll


Roger-luo:
this is current a bit hard to implement because Python uses `next` which has side effect, thus const prop won't be able to "just" fold the loop branches. We need extra analysis to rewrite `next` into some more Julia-like pure functions like `iterate` so the states are passed in immutable manner.

PR #390, opened 2025-05-02
Iter stmt should be non-pure

kaihsin:
This PR address #389.

Stmt `Iter` to get iterator should not be pure otherwise constfold will cached the iterator itself which is not reusable.

### DIFF
--- a/src/kirin/dialects/py/iterable.py
+++ b/src/kirin/dialects/py/iterable.py
@@ -26,7 +26,7 @@ PyRangeIterType = types.PyClass(type(iter(range(0))))
 class Iter(ir.Statement):
     """This is equivalent to `iter(value)` in Python."""
 
-    traits = frozenset({ir.Pure()})
+    traits = frozenset({})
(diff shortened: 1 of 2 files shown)


cduck:
Question for @Roger-luo: Does the iter statement being non-pure mean that any function containing a for-loop automatically become non-pure for optimization purposes?  Will this artificially limit compiler optimization opportunities?

Roger-luo:
@cduck  Yes, this is because in `basic`-lang the IR only has `cf.Branch` and `cf.ConditionalBranch`, see also #188 so we need to write explicitly how iterators are unpacked and updated, a.k.a `Iter`, `Next` with an explicit state. We will need to have effect analysis (which we don't have) to prove the state won't escape else where then we will be able to trace the state and fold it. But if we use MLIR/LLVM in the future I would imagine that would handle a large portion of these cases.

Also note that if we let frontend user use `scf` and have a rewrite to `cf` a lot of constant loop will be trivial to analyze.

PR #404, opened 2025-05-13
Rewrite rule from scf to cf

weinbe58:
this PR implements the rewrite rule from SCF to CF dialect.

This rule must be applied in a fixed point fashion and works by walking a ssacfg region looking for `scf.For` and `scf.IfElse`. when if finds these statements it inserts their body blocks and uses `cf` to create the appropriate cfg for those statements. This only implements it for one level of nesting; hence, it should be applied in a fixed-point fashion.

### DIFF
--- a/src/kirin/dialects/scf/scf2cf.py
+++ b/src/kirin/dialects/scf/scf2cf.py
@@ -5,8 +5,131 @@ from ...rewrite.abc import RewriteRule, RewriteResult
 
 class ScfToCfRule(RewriteRule):
 
-    def rewrite_ssacfg(self, node: ir.Region):
+    def rewrite_ifelse(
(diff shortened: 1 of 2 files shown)


src/kirin/dialects/func/stmts.py
```diff
@@ -51,6 +51,8 @@ class Function(ir.Statement):
     """The signature of the function at declaration."""
     body: ir.Region = info.region(multi=True)
     """The body of the function."""
+    result: ir.ResultValue = info.result(MethodType)
+    """The result of the function."""
```
weinbe58, line 55: Weird I have no idea where this commit came from

weinbe58: oh nvm I see

weinbe58:
fyi I think this PR is currently blocked by https://github.com/QuEraComputing/kirin/issues/406

Issue #416, opened 2025-05-21
Fold pass removes if / elif when it shouldn't


david-pl:
MWE:

```python
from kirin.prelude import structural_no_opt
from kirin.passes import Fold

@structural_no_opt
def main(n: int):
    x = 0

    if x == n:
        x += 1
    elif x == n+1:
        x += 3
    
    return x

main.print()

Fold(main.dialects)(main)

main.print()
```

Output:

```python
func.func main(!py.int) -> !Any {
  ^0(%main_self, %n):
  │ %x = py.constant.constant 0 : !py.int
  │      func.return %x
} // func.func main
```

Things work correctly if you replace the `elif` by another `if`, but it's not technically identical since `x` is updated in-place which might lead to the second `if` to be `True.`.

david-pl:
Oh, I think the issue is actually not with the `Fold` pass. This is working as intended, kind of.

The issue is that as soon as there's an `elif`, the SSA value doesn't get updated properly anymore.

```python
@structural_no_opt
def main(n: int):
    x = 0
    if x == n:
        x = 1
    elif x == n+1:
        x = 3
    return x

main.print()
```

Output:

```python
func.func @main(n : !py.int) -> !Any {
  ^0(%main_self, %n):
  │ %x = py.constant.constant 0 : !py.int
  │ %0 = py.cmp.eq(lhs=%x, rhs=%n : !py.int) : !py.bool
  │      scf.if %0 {
  │      │ ^1(%1):
  │      │ │ %x_1 = py.constant.constant 1 : !py.int
  │      │ │        scf.yield 
  │      } else {
  │      │ ^2(%5):
  │      │ │ %2 = py.constant.constant 1 : !py.int
  │      │ │ %3 = py.binop.add(%n : !py.int, %2) : ~T
  │      │ │ %4 = py.cmp.eq(lhs=%x, rhs=%3) : !py.bool
  │      │ │      scf.if %4 {
  │      │ │      │ ^3(%6):
  │      │ │      │ │ %x_2 = py.constant.constant 3 : !py.int
  │      │ │      │ │        scf.yield 
  │      │ │      } -> purity=False
  │      │ │      scf.yield 
  │      } -> purity=False
  │      func.return %x
} // func.func main
```

The value is just assigned to `%0` which is unused.

david-pl:
Slowly digging further into this: the bug is actually triggered whenever there is an assignment to a variable from the outer scope in only one of the branches. Consider, for example

```python
[...]

PR #417, opened 2025-05-26
Fix bug that ignored defs in if when missing from else

david-pl:
@Roger-luo this fixes the bug in #416 when there's a definition in the `if` path, but not in the `else` path.

However, the issue with the `elif` persists since the definition from the nested if isn't propagated upwards properly.

In code:

This fails on main, but works on this branch

```python
@kernel
def main(n: int):
    x = 0

    if x == n:
        x = 1
    else:
        y = 2  # noqa: F841

    return x

main.print()
```

This fails both on main and this branch:
```python
@kernel
def main_elif(n: int):
    x = 0

    if x == n:
        x = 3
    elif x == n + 1:
        x = 4

    return x

main_elif.print()
```

(or, equivalently)
```python
@kernel
def main_elif(n: int):
    x = 0

    if x == n:
        x = 3
    else:
        if x == n + 1:
            x = 4

    return x

```

We somehow need to lower the bodies first in order to obtain the nested definitions in them, but I'm not sure how exactly.

### DIFF
--- a/src/kirin/dialects/scf/lowering.py
+++ b/src/kirin/dialects/scf/lowering.py
@@ -10,6 +10,22 @@ from ._dialect import dialect
 @dialect.register
 class Lowering(lowering.FromPythonAST):
 
+    @staticmethod
+    def _frame_or_any_parent_has_def(frame, name) -> ir.SSAValue | None:
+        # NOTE: this recursively checks all parents of the current frame for the
+        # def. Required for nested if statements that e.g. assign to variables
+        # defined in outer scope
+        if frame is None:
+            return None
+
+        if name in frame.defs:
+            value = frame.get(name)
+            if value is None:
+                raise lowering.BuildError(f"expected value for {name}")
+            return value
+
+        return Lowering._frame_or_any_parent_has_def(frame.parent, name)
+
     def lower_If(self, state: lowering.State, node: ast.If) -> lowering.Result:
         cond = state.lower(node.test).expect_one()
         frame = state.current_frame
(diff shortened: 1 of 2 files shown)


david-pl:
pyright failure seems to be unrelated, I think.

Roger-luo (changes requested):
actually why do we remove `node.orelse` here?

david-pl:
There we go: the issue that kept the `elif` from working was that the `if` nested in the `else` didn't check definitions present in its parent frame.

I added

```python
if frame.parent is not None:
            frame.defs.update(frame.parent.defs)
```
 and now everything seems to work.

The example from before

```python
@kernel
def main_elif(n: int):
    x = 0

    if x == n:
        x = 3
    elif x == n + 1:
        x = 4

    return x

main_elif.print()
```

now prints

```python
func.func @main_elif(n : !py.int) -> !Any {
  ^0(%main_elif_self, %n):
  │   %x = py.constant.constant 0 : !py.int
  │   %0 = py.cmp.eq(lhs=%x, rhs=%n : !py.int) : !py.bool
  │ %x_1 = scf.if %0 {
  │        │ ^1(%1):
  │        │ │ %x_2 = py.constant.constant 3 : !py.int
  │        │ │        scf.yield %x_2
  │        } else {
  │        │ ^2(%5):
  │        │ │   %2 = py.constant.constant 1 : !py.int
  │        │ │   %3 = py.binop.add(%n : !py.int, %2) : ~T
  │        │ │   %4 = py.cmp.eq(lhs=%x, rhs=%3) : !py.bool
  │        │ │ %x_3 = scf.if %4 {
  │        │ │        │ ^3(%6):
  │        │ │        │ │ %x_4 = py.constant.constant 4 : !py.int
  │        │ │        │ │        scf.yield %x_4
  │        │ │        } else {
  │        │ │        │ ^4(%7):
  │        │ │        │ │ scf.yield %x
  │        │ │        } -> purity=False
  │        │ │        scf.yield %x_3
  │        } -> purity=False
  │        func.return %x_1
} // func.func main_elif
```

That looks correct to me!

@Roger-luo ready for another round of review!

Roger-luo:
I think this breaks some of the test case?

david-pl:
Ah, damn, I need to take another look then.

Roger-luo:
LGTM!

Roger-luo:
let me also try to backport it to 0.17 since that's what we currently depends on

PR #421, opened 2025-06-03
RFC: support overwriting existing method

Roger-luo:
this PR add support of overwriting existing method by directly updating the existing method object in (current Python not Kirin) scope if the name matches. We limit the compiler not to update existing method because this may cause unexpected behaviour because we won't actually update existing compiled code. However, after thinking a bit more, I think this is still ok because we can keep the same method object but update its body. 

The caveat is we need to retrigger the whole compilation of call the caller of this method, which is probably still ok? I didn't implement the [entire world age mechanism in Julia](https://d1wqtxts1xzle7.cloudfront.net/73563045/3428275-libre.pdf?1635141917=&response-content-disposition=inline%3B+filename%3DWorld_age_in_Julia_optimizing_method_dis.pdf&Expires=1749052521&Signature=OpvF9l3YO6ZRxHylvBFSMIH0O1Fl6AO2nyg1EZuD-5SWpZZw8E50yy7mgLcGMKhCwTWm6414MhLU9BwolYYnId-ichrBKTkxiDH3MPp050NFtXtmuGQ0Hxnc2wgeDgvKJPiNfVLydlPQLu7daEWw9uB4J2rAzb42m37YBOkfVvdjwvBHg8ysLncdDs-rc15mXKCmvsAYTFl7zRdaUxxhgphNYO~SOdNLB2QWxPpA8SVu3p-HGGitEzyjhtBJi9aQzH0xgBStX6~l2DIyMLEwlhYwjByrw0~moMhKWmOpRELyTfQdYeOxXD39YcztERPOFCE9dTHgzFv0fX-tMaDfkA__&Key-Pair-Id=APKAJLOHF5GGSLRBV4ZA) but technically every time you define a new method it should trigger an eval -ish process that recompiles the related methods and thus creates a new world age. So I'd like to see if there are volunteers willing to take a second look and play with this PR a bit before moving forward.

## Motivation

currently when supporting interactive environments, such as a notebook, it is quite cumbersome to use notebook as a dev environment like how one codes Python natively because we cannot update the code of an existing `Method`.


Roger-luo:
I'm merging this first so people can play with it.

Issue #419, opened 2025-05-29
Typos in Pauli algebra tutorial


ehua7365:
`result.RewriteResult` needs to be replaced with `abc.RewriteResult` for it to work.

Markdown with [Fold] link link.

Should be $2\times 2$ matrix not $2x2$ matrix.

PR #423, opened 2025-06-05
Update Pauli example

david-pl:
The problem with the link to the Fold pass is that there's a problem with the nav generation, I think. The page of the fold pass is just empty: https://queracomputing.github.io/kirin/latest/reference/kirin/passes/fold/

Ref #419


david-pl:
Ah, I see. Class without docstrings don't get picked up in the build.

PR #441, opened 2025-07-16
Testing tools to compare Methods against lists of Statements

ehua7365:
Add the `kirin.test` module containing useful assert functions for writing tests that includes
- `assert_structurally_same` for comparing `Method` against a list of `Statements`
- `assert_statements_same` for comparing `Statement` to other `Statement` objects
These methods compare the attributes of statements, as well as the arguments of the statements recursively.
Comparing the arguments can be turned off setting `check_args=False`.
Also has a `debug` option that prints out the statements in the method one by one until it fails, so the user can identify on which statement the Method fails to match the list of Statements.

### DIFF
--- a/src/kirin/test/statements.py
+++ b/src/kirin/test/statements.py
@@ -3,6 +3,7 @@
 from typing import List
 
 from kirin.ir import Method, Statement
+from kirin.dialects import func
 
 
 def assert_statements_same(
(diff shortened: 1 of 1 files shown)


kaihsin:
Its recommended not to use `test` as any submodule and python script name, as it will confuse pytest.

Roger-luo:
I'll do the rename separately!

Issue #444, opened 2025-07-22
`ScfToCFRule` Not compatible with v0.17.0


weinbe58:
Specifically this check based on the traits: [here](https://github.com/QuEraComputing/kirin/blob/95b478e7a3c22cb3550a2f437bfb531983aa6ab4/src/kirin/dialects/scf/scf2cf.py#L180)

We can probably just remove these checks since the `scf` dialect should contain SSACFG regions.

PR #449, opened 2025-07-24
Fix type inference of loop variable on v0.17

david-pl:
**NOTE**: this is a **backport only fix**, i.e. it's a patch for v0.17 only (the issue is no longer present on v0.18+).

This is a small fix that causes a lot of trouble downstream. It was easier to fix here than adding workarounds in bloqade-circuit. I hope that's okay.

As far as I can tell, the only thing we "lose" are a couple of tests (which I added here) and it makes future fixes that touch the same code more difficult to backport (but it's only three lines I touched).

### DIFF
--- a/src/kirin/dialects/scf/scf2cf.py
+++ b/src/kirin/dialects/scf/scf2cf.py
@@ -175,11 +175,7 @@ class ScfToCfRule(RewriteRule):
         return RewriteResult(has_done_something=has_done_something)
 
     def rewrite_Statement(self, node: ir.Statement) -> RewriteResult:
-        if (
-            isinstance(node, (For, IfElse))
-            or not node.has_trait(ir.HasCFG)
-            and not node.has_trait(ir.SSACFG)
-        ):
+        if isinstance(node, (For, IfElse)):
             # do not do rewrite in scf regions
             return RewriteResult()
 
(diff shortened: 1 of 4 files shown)


david-pl:
@Roger-luo I'm not sure why CI is failing here, but it seems it's already failing on v0.17.12 as is (without the changes from this PR).

Roger-luo (approved):

src/kirin/dialects/scf/scf2cf.py
```diff
@@ -175,11 +175,7 @@ def rewrite_ssacfg(self, node: ir.Region):
         return RewriteResult(has_done_something=has_done_something)
 
     def rewrite_Statement(self, node: ir.Statement) -> RewriteResult:
-        if (
-            isinstance(node, (For, IfElse))
-            or not node.has_trait(ir.HasCFG)
-            and not node.has_trait(ir.SSACFG)
```
Roger-luo, line 181: can you elaborate how does this fix the issue?

david-pl: This is unrelated, it's issue #444. The scf2cf is not compatible with v0.17 if we leave those in, in that the traits `HasCFG` and `SSACFG` just don't exist in v0.17, so I removed those.

Sorry for mixing those commits. You could consider rebasing the branch without squashing so we have them separately.

Roger-luo: I see that's ok, this is 0.17 specific. I was just wondering.

PR #460, opened 2025-08-06
fixing bug in which purity wasn't propagated properly

weinbe58:
The following function is definitely pure:

```python
    @structural_no_opt
    def test_func(src: ilist.IList[float, Any]):

        def inner(i: int):
            if src[i] < 0:
                return 0.0
            elif src[i] < 1.0:
                return 1.0
            else:
                return 2.0

        return ilist.map(inner, ilist.range(len(src)))
```

but there is a bug in const prop when you have return values. The propagation of frame purity was not occurring, making the constant property of this function impure, which created other issues regarding folding and DCE.

### DIFF
--- a/test/dialects/scf/test_constprop.py
+++ b/test/dialects/scf/test_constprop.py
@@ -232,3 +232,25 @@ def test_purity_7():
     frame, ret = prop.run(test_func)
 
     assert frame.frame_is_not_pure, "function should not be pure"
+
+
+def test_purity_8():
+
+    @structural_no_opt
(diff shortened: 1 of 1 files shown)


weinbe58:
> Does constprop run before the PickIfElse?
> 
> What if something like
> 
> if true: .... else: ....

PickIfElse will only use the hints so if there is no hint you can't pick which branch.

weinbe58:
Yes this will still be pure because when you hit the `IfElse` it will only run the analysis on the True block. The case I'm editing is the case where the `cond` is `Unkown()`

PR #462, opened 2025-08-07
Revert "fixing bug in which purity wasn't propagated properly"

weinbe58:
Reverts QuEraComputing/kirin#460

Turns out that semantically you have to make the IfElse body not pure if there is a return value because of a weird interaction with ConstantFold. Namely the fact that the return value of the if statement has no uses, and so it will be deleted.


kaihsin:
temporary revert for correctness of program, but need better way to deal with early termination/return from child frames (scf bodies)

kaihsin (approved):

Issue #463, opened 2025-08-07
Bug when inlining `scf.IfElse` function with `Return` in both bodies.


weinbe58:
Take the following example in v0.17.16:

```python
from kirin.rewrite import Inline, Walk
from kirin.passes import Fold
from kirin.prelude import structural


@structural
def sign(cond):
    if cond:
        return 1
    else:
        return -1

@structural
def main(cond):
    return [sign(cond)]


Walk(Inline(lambda x:True)).rewrite(main.code)
Fold(structural)(main)

main.print()
```

gives

```mlir
func.func main(!Any) -> !Any {
  ^0(%main_self, %cond):
  │      scf.if %cond {
  │      │ ^1(%cond_1):
  │      │ │ %2 = py.constant.constant 1 : !py.int
  │      │ │      func.return %2
  │      } else {
  │      │ ^2(%cond_2):
  │      │ │ %3 = py.constant.constant -1 : !py.int
  │      │ │      func.return %3
  │      } -> purity=False
  │ %0 = func.const.none() : !py.NoneType
  │ %1 = py.ilist.new(values=(%0)){elem_type=!Any} : !py.IList[!Any, Literal(1,int)]
  │      func.return %1
} // func.func main
```

Which is incorrect,  on a few levels:

1. The ilist being created contains None which is incorret
2. The resulting function will terminate too early compared to the original one

This seems to be the cause of this issue: https://github.com/QuEraComputing/bloqade-shuttle/issues/47

weinbe58:
I think the reason why this doesn't work is because inlining assumes any `Return` statements will be on the control flow graph of the ssacfg, but here this is not the case.

Roger-luo:
so we should fix lowering and ban return in the regions of scf and only allow `yield` as terminator

PR #464, opened 2025-08-08
Pass that applies RewriteRule to all functions on a call graph. 

weinbe58:
This PR is mostly needed in Kirin 0.17, as some bugs prevent the usage of inlining, for example, issue #463. 

In that particular case, we need to first remove all SCF statements in all methods on the call graph before we can safely inline subroutines. 

This pass could also save on some compilation time by applying rewrites across subroutines instead of having to inline them first.

### DIFF
--- a/src/kirin/passes/callgraph.py
+++ b/src/kirin/passes/callgraph.py
@@ -45,19 +45,22 @@ class CallGraphPass(passes.Pass):
     rule: RewriteRule
     """The rule to apply to each function in the call graph."""
 
-    def methods_on_callgraph(self, mt: ir.Method) -> set[ir.Method]:
+    @staticmethod
+    def methods_on_callgraph(mt: ir.Method) -> set[ir.Method]:
 
         callees = {mt}
         stack = [mt]
 
         while stack:
             current_mt = stack.pop()
+
+            if current_mt in callees:
+                continue
+
             for stmt in current_mt.callable_region.walk():
-                if isinstance(stmt, Invoke):
-                    callee = stmt.callee
-                    if callee not in callees:
-                        callees.add(callee)
-                        stack.append(callee)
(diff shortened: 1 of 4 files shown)


src/kirin/passes/callgraph.py
```diff
@@ -0,0 +1,85 @@
...
+                new_mt = original_mt
+            else:
+                new_mt = original_mt.similar()
+            result = self.rule.rewrite(new_mt.code).join(result)
+            mt_map[original_mt] = new_mt
+
+        if result.has_done_something:
+            for _, new_mt in mt_map.items():
+                rewrite.Walk(ReplaceMethods(mt_map)).rewrite(new_mt.code)
+                passes.Fold(self.dialects, no_raise=self.no_raise)(new_mt)
```
kaihsin, line 83: I think this should use `new_mt.dialects` instead of `self.dialects`?

kaihsin: Then in that case the program it produce is invalid because it contains
dialects that it's not recognized by the dialect group that method is
carrying.

kaihsin: In general I agree, but since you are doing similar() in Pass. If thats the case, I don't think something like similar should be within a pass

kaihsin: I think a pass should be a bundle of rewrite rules, and should be always mutate the input method. something like similar should live outside a pass to avoid this confusion

weinbe58: But how are you supposed to change the dialect groups of the subroutines? You would need to make a special pass to traverse the call graph do the copy and reassign dialects which is basically what this is doing

kaihsin:
oh, ok. those are similar for the invoked methods. I mistaken you also similar the input. then I have no question/issue on this.

Issue #471, opened 2025-08-11
Adding vmath dialect


kaihsin:
Adding vectorized math dialect that takes ilist and return ilist

kaihsin:
completed in #473

PR #473, opened 2025-08-12
This PR add vmath dialect

kaihsin:
Add vectorized math dialect. address #471


pyproject.toml
```diff
@@ -7,6 +7,8 @@ dependencies = [
     "rich>=13.7.1",
     "beartype>=0.17.2",
     "typing_extensions>=4.11.0",
+    "numpy>=2.2.6",
+    "scipy>=1.15.3",
```
Roger-luo, line 11: can we make this optional and making `vmath` an extra?

Roger-luo (changes requested):
for a future note, we should think about supporting array after upgrading the type system (it's cubersome to support custom generic type with custom inference rule right now)

I think we should not have this enabled by default and this dialect should be an extra given it will need to depend on numpy and scipy (can we also separate the numpy and scipy maybe?)

pyproject.toml
kaihsin: moved this as an optional

PR #479, opened 2025-08-19
Adding fallback when attribute isn't hashable

### LINKED ISSUE #472: CommonSubexpressionElimination tries to hash non-hashable runtime values
Specifically, CommonSubexpressionElimination fails to optimize a kernel that contains `slice` objects because `slice` does not support hash (this was a design choice in Python to avoid a misunderstanding like `my_dictionary[1:3]`).

I discovered this bug in UnrollPass from #405 which uses CommonSubexpressionElimination.

weinbe58:
closes #472

### DIFF
--- a/src/kirin/rewrite/cse.py
+++ b/src/kirin/rewrite/cse.py
@@ -36,12 +36,9 @@ class Info:
         return self._hash
 
     def __eq__(self, other: object) -> bool:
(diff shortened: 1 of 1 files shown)


kaihsin (approved):

kaihsin:
LGTM

Issue #140, opened 2025-01-12
peephole optimization


Roger-luo:
some simple optimization can be done right now

- [ ] similar term merging `add(%a, %a)` -> `mul(2, %a)`, `add(%a, mul(2, %a))` -> `mul(3, %a)` etc.

PR #486, opened 2025-08-26
Dl/peephole optimize

zhenrongliew:
adds Peephole optimization rewrites and test cases. For issue https://github.com/QuEraComputing/kirin/issues/140

### DIFF
--- a/src/kirin/rewrite/peephole.py
+++ b/src/kirin/rewrite/peephole.py
@@ -1,0 +1,42 @@
+from dataclasses import dataclass
+
+from kirin import ir
+from kirin.dialects import py
+from kirin.rewrite.abc import RewriteRule, RewriteResult
+
+
+@dataclass
+class PeepholeOptimize(RewriteRule):
+
+    def rewrite_Statement(self, node: ir.Statement) -> RewriteResult:
+
+        if isinstance(node, py.binop.Add):
+            #   add(%a, %a) -> mul(2, %a)
+            if node.lhs == node.rhs:
+                print("-> Enter equal")
+                x = py.Constant(2)
+                x.insert_before(node)
+                new_stmt = py.binop.Mult(x.result, node.rhs)
+                node.replace_by(new_stmt)
+                return RewriteResult(has_done_something=True)
+
+            #   add(mul(2, %a), %a) -> mul(3, %a)
+            elif isinstance(mult_node := node.lhs.owner, py.binop.Mult):
+                if isinstance(const_node := mult_node.lhs.owner, py.Constant):
+                    print("E")
+                    x = const_node.value.unwrap()
+                    const_node.replace_by(py.Constant(x + 1))
+                    node.replace_by(py.binop.Mult(mult_node.lhs, node.rhs))
+                    mult_node.delete()
+                    return RewriteResult(has_done_something=True)
+
+            #   add(%a, mul(2, %a)) -> mul(3, %a)
+            elif isinstance(mult_node := node.rhs.owner, py.binop.Mult):
+                if isinstance(const_node := mult_node.lhs.owner, py.Constant):
+                    x = const_node.value.unwrap()
+                    const_node.replace_by(py.Constant(x + 1))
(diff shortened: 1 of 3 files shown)


src/kirin/rewrite/peephole.py
```diff
@@ -0,0 +1,42 @@
...
+
+@dataclass
+class PeepholeOptimize(RewriteRule):
+
+    def rewrite_Statement(self, node: ir.Statement) -> RewriteResult:
+
+        if isinstance(node, py.binop.Add):
+            #   add(%a, %a) -> mul(2, %a)
+            if node.lhs == node.rhs:
+                print("-> Enter equal")
```
weinbe58, line 16: don't forget to remove any debug prints!

src/kirin/rewrite/peephole.py
```diff
@@ -0,0 +1,42 @@
...
+                x = py.Constant(2)
+                x.insert_before(node)
+                new_stmt = py.binop.Mult(x.result, node.rhs)
+                node.replace_by(new_stmt)
+                return RewriteResult(has_done_something=True)
+
+            #   add(mul(2, %a), %a) -> mul(3, %a)
+            elif isinstance(mult_node := node.lhs.owner, py.binop.Mult):
+                if isinstance(const_node := mult_node.lhs.owner, py.Constant):
+                    print("E")
```
weinbe58, line 26: same here

src/kirin/rewrite/peephole.py
```diff
@@ -0,0 +1,42 @@
...
+                    x = const_node.value.unwrap()
+                    const_node.replace_by(py.Constant(x + 1))
+                    node.replace_by(py.binop.Mult(mult_node.lhs, node.rhs))
+                    mult_node.delete()
+                    return RewriteResult(has_done_something=True)
+
+            #   add(%a, mul(2, %a)) -> mul(3, %a)
+            elif isinstance(mult_node := node.rhs.owner, py.binop.Mult):
+                if isinstance(const_node := mult_node.lhs.owner, py.Constant):
```
weinbe58, line 35: nit-pick: you could remove this nested if statement using an `and`

src/kirin/rewrite/peephole.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+@dataclass
+class PeepholeOptimize(RewriteRule):
+
+    def rewrite_Statement(self, node: ir.Statement) -> RewriteResult:
+
+        if isinstance(node, py.binop.Add):
+            #   add(%a, %a) -> mul(2, %a)
+            if node.lhs == node.rhs:
```
weinbe58, line 15: This is fine but each ssa value is a unique object so you really should use `is` to make this more clear. 
```suggestion
            if node.lhs is node.rhs:
```

src/kirin/rewrite/peephole.py
```diff
@@ -0,0 +1,42 @@
...
+                    x = const_node.value.unwrap()
+                    const_node.replace_by(py.Constant(x + 1))
+                    node.replace_by(py.binop.Mult(mult_node.lhs, node.rhs))
+                    mult_node.delete()
+                    return RewriteResult(has_done_something=True)
+
+            #   add(%a, mul(2, %a)) -> mul(3, %a)
+            elif isinstance(mult_node := node.rhs.owner, py.binop.Mult):
+                if isinstance(const_node := mult_node.lhs.owner, py.Constant):
+                    x = const_node.value.unwrap()
```
weinbe58, line 36: This value may not neccesarily be a float or an integer you should make sure to check this before adding `1` to the value.

weinbe58 (changes requested):
I think overall it looks like a good first pass but one big issue is that these optimization should only be applied to the built-in types `float` and `int`. I would suggest adding a gaurd at the top of the `rewrite_Statement` that does something like:

```python
result_type = node.result.type
if not result_type.is_subseteq(types.Float) or not result_type.is_subseteq(types.Int):
    return RewrireResult(has_done_something=False)
```

weinbe58 (approved):
LGTM

Issue #468, opened 2025-08-11
Lowering bug in `scf.IfElse`


weinbe58:
mwe: 

```python
from kirin.prelude import structural

@structural(typeinfer=True, fold=False)
def main(n: int):
    c = 1.0
    if n <= 0:
        return 0.0
    else:
        c = 2.0

    return c

main.print()
```

gives

```mlir
func.func @main(n : !py.int) -> !Union[Literal(0.0,float), Literal(1.0,float)] {
  ^0(%main_self, %n):
  │ %c = py.constant.constant 1.0 : !py.float
  │ %0 = py.constant.constant 0 : !py.int
  │ %1 = py.cmp.lte(lhs=%n : !py.int, rhs=%0) : !py.bool
  │      scf.if %1 {
  │      │ ^1(%3):
  │      │ │ %2 = py.constant.constant 0.0 : !py.float
  │      │ │      func.return %2
  │      } else {
  │      │ ^2(%4):
  │      │ │ %c_1 = py.constant.constant 2.0 : !py.float
  │      │ │        scf.yield 
  │      } -> purity=False
  │      func.return %c
```

which is not correct, the if statement should yield a value for `c`. 

EDIT: This is only an issue on previous versions as we have banned return statements in #495

PR #487, opened 2025-08-27
Dl/scf lowering miss yield

zhenrongliew:
Fixes issue #468


zhenrongliew:
@Roger-luo yes, this is backportable to 0.17.

Roger-luo:
@zhenrongliew can we backport this to 0.17? thanks!

PR #492, opened 2025-09-15
Serializer framework for Kirin

zhenrongliew:
Serializer framework. With base Serializer, JSONSerializer, and BinarySerializer.

### DIFF
--- a/src/kirin/ir/nodes/region.py
+++ b/src/kirin/ir/nodes/region.py
@@ -261,7 +261,7 @@ class Region(IRNode["Statement"]):
         """
         if context is None:
             context = {}
-
+        context[self] = other
         if len(self.blocks) != len(other.blocks):
             return False
 
--- a/src/kirin/ir/nodes/stmt.py
+++ b/src/kirin/ir/nodes/stmt.py
@@ -551,8 +551,12 @@ class Statement(IRNode["Block"]):
         Returns:
             bool: True if the IRNode is structurally equal to the other.
         """
+        if self is other:
+            return True
+
         if context is None:
             context = {}
+        context[self] = other
 
         if self.name != other.name:
             return False
@@ -577,6 +581,12 @@ class Statement(IRNode["Block"]):
             and other.parent is not None
             and context.get(self.parent) != other.parent
         ):
+            print(id(self.parent))
+            self.parent.print()
+            print(id(other.parent))
+            other.parent.print()
+            for c, v in context.items():
+                print("id:", id(c), "value:", id(v))
             return False
 
         if not all(


src/kirin/dialects/func/attrs.py
```diff
@@ -38,3 +39,14 @@ def print_impl(self, printer: Printer) -> None:
         printer.print_seq(self.inputs, delim=", ", prefix="(", suffix=")")
         printer.plain_print(" -> ")
         printer.print(self.output)
+
+    def serialize(self) -> dict[str, Any]:
+        # return the plain inner payload ONLY; global Serializer will encode nested
+        return {"inputs": [a for a in self.inputs], "output": self.output}
```
kaihsin, line 45: shouldn't it be `[a for a in self.inputs.serialize()]` and `self.output.serialize()`?

src/kirin/dialects/func/attrs.py
```diff
@@ -38,3 +39,14 @@ def print_impl(self, printer: Printer) -> None:
...
+
+    @classmethod
+    def deserialize(cls, data: Dict[str, Any]) -> "Signature":
+        # accept the plain inner payload produced by serialize()
+        inputs = tuple(data.get("inputs", ()))
+        output = data.get("output")
```
kaihsin, line 51: same goes here, shouldn't it be `TypeLatticeElem.deserialize(input)` and `TypeLatticeElem.deserialize(output)`

src/kirin/dialects/ilist/runtime.py
```diff
@@ -90,3 +91,21 @@ def print_impl(self, printer: Printer) -> None:
...
+
+    @classmethod
+    def deserialize(cls, data: dict[str, Any]) -> "IList":
+        items = []
+        for x in data.get("data", []):
+            items.append(x)
+        # 'elem' is expected to be already decoded by the global Serializer.
+        # Do not call .deserialize() here (that caused the AttributeError).
+        elem = data.get("elem", types.Any)
+        return IList(items, elem=elem)
```
kaihsin, line 111: I feel like something is a bit off here @Roger-luo

kaihsin: maybe we need to pass a global context in deserialize/serialize?

src/kirin/ir/attrs/py.py
```diff
@@ -51,3 +53,15 @@ def print_impl(self, printer: Printer) -> None:
 
     def unwrap(self) -> T:
         return self.data
+
+    def serialize(self) -> dict[str, Any]:
+        return {"data": self.data}
```
kaihsin, line 58: need to check if self.data has Mixin, if it is, call serialize

src/kirin/serialization/base/registry.py
```diff
@@ -0,0 +1,343 @@
...
+    def decode(self, data):
+        if data.get("kind") != "runtime":
+            raise ValueError("Invalid runtime data for decoding.")
+
+        style = data.get("style")
+        if style not in RUNTIME_DECODE_LOOKUP:
+            raise ValueError(f"No registered decode method for style {style}.")
+
+        return RUNTIME_DECODE_LOOKUP[style](self, data.get("data"))
+
```
kaihsin, line 213: Do we still need this? we just ask user to implement runtime that has the SerializeMixin

src/kirin/serialization/base/registry.py
```diff
@@ -0,0 +1,343 @@
...
+            "type": self.encode(obj.type) if obj.type else None,
+        }
+
+    def _decode_Literal(self, content_data: dict):
+        value = content_data["value"]
+        typ = content_data.get("type")
+        if typ:
+            typ = self.decode(typ)
+
+        return types.Literal(data=value, datatype=typ)
```
kaihsin, line 338: Can we make Attribute serializer generic to include both PyAttr and TypeAttr?

src/kirin/serialization/binaryserializer.py
```diff
@@ -0,0 +1,17 @@
...
+    def __init__(self, types: list[type] = []):
+        super().__init__(types=types)
+
+    def encode(self, obj) -> bytes:
+        data = super().encode(obj)
+        return pickle.dumps(data, protocol=pickle.HIGHEST_PROTOCOL)
+
+    def decode(self, payload: bytes) -> Any:
+        data = pickle.loads(payload)
+        return super().decode(data)
```
kaihsin, line 17: I don't think we should use pickle.

test/serialization/test_jsonserializer.py
```diff
@@ -0,0 +1,75 @@
...
+    serializer = JSONSerializer()
+    s1 = serializer.encode(main)
+    s2 = serializer.encode(main)
+    assert s1 == s2
+
+
+test_to_json1()
+test_to_json2()
+test_to_json3()
+test_deterministic()
```
kaihsin, line 75: remember to remove these from test

kaihsin (changes requested):
btw, linting CI failed

src/kirin/dialects/func/attrs.py
```diff
@@ -40,13 +40,18 @@ def print_impl(self, printer: Printer) -> None:
         printer.plain_print(" -> ")
         printer.print(self.output)
 
-    def serialize(self) -> dict[str, Any]:
-        # return the plain inner payload ONLY; global Serializer will encode nested
-        return {"inputs": [a for a in self.inputs], "output": self.output}
+    def serialize(self, serializer) -> dict[str, Any]:
```
Roger-luo, line 42:
```suggestion
    def serialize(self, serializer) -> dict[str, Any]:
```

I think we should have either a base class or a protocal (pure type hint) for the `serializer`, we should not have any none-type-hinted code in this code base

src/kirin/serialization/base/registry.py
```diff
@@ -0,0 +1,147 @@
...
+    def wrapper(func):
+        import warnings
+
+        warnings.warn(
+            "runtime_register_encode is deprecated and will be removed; "
+            "prefer the new Serializer API.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        RUNTIME_ENCODE_LOOKUP[obj_type.__name__] = func
```
weinbe58, line 64: using the name might be problematic here since an object's name might happen multiple times between different dialects. 

This should at least throw an error if there is a name collision.

src/kirin/ir/attrs/py.py
```diff
@@ -51,3 +55,19 @@ def print_impl(self, printer: Printer) -> None:
 
     def unwrap(self) -> T:
         return self.data
+
+    def serialize(self, serializer: "Serializer") -> dict[str, Any]:
```
Roger-luo, line 59: what if I have the following attribute

```python
class MyAttr(Data[T]):
     data: T
     type: SomeCustomType
```

this serialization target format cannot recover the original attribute information, we can solve this by also serializing the name of an attribute.

Roger-luo (approved):
LGTM

test/serialization/test_jsonserializer.py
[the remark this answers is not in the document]
zhenrongliew: Added new test for kernel to kernel calls. Edited old test to include some ilist, slices, and tuples.
To pass test, needed to edit `is_structurally_equal` in stmt, and define new ones to PyAttr, Signature, Method, DialectGroup.

Roger-luo (approved):

PR #499, opened 2025-09-18
Make Literal compare as == intead of just id

kaihsin:
As per discussion, 

we are making this Literal comparison `is_equal` to be == instead of just compare id, since it contain runtime. 
Another issue will be created to refactor this to make it more performant


Roger-luo:
can you also backport this in a new PR so I can release a new version on 0.17?

PR #505, opened 2025-09-25
Adding rewrite to flatten sum of two ilist.New statements

weinbe58:
This is being ported from Bloqade-circuit. This i needed in a few fold passes.

### DIFF
--- a/src/kirin/dialects/ilist/rewrite/flatten_add.py
+++ b/src/kirin/dialects/ilist/rewrite/flatten_add.py
@@ -1,0 +1,21 @@
+from dataclasses import dataclass
+
+from kirin import ir
+from kirin.dialects import py, ilist
+from kirin.rewrite.abc import RewriteRule, RewriteResult
+
+
+@dataclass
+class FlattenAddOpIList(RewriteRule):
+
+    def rewrite_Statement(self, node: ir.Statement) -> RewriteResult:
+        if (
+            not isinstance(node, py.binop.Add)
+            or not isinstance(lhs := node.lhs.owner, ilist.New)
+            or not isinstance(rhs := node.rhs.owner, ilist.New)
+        ):
+            return RewriteResult()
+
+        node.replace_by(ilist.New(values=lhs.values + rhs.values))
+
+        return RewriteResult(has_done_something=True)


src/kirin/dialects/ilist/rewrite/flatten_add.py
```diff
@@ -0,0 +1,40 @@
+from kirin import ir
+from kirin.analysis import const
+from kirin.dialects import py, ilist
+from kirin.rewrite.abc import RewriteRule, RewriteResult
+
+
+class FlattenAdd(RewriteRule):
+
+    def rewrite_Statement(self, node: ir.Statement) -> RewriteResult:
+        if (
```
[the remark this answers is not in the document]
weinbe58: it if is not add -> skip
if lhs is not an ilist type -> skip 
if rhs is not an ilist type -> skip 

Put those into an or because if any are true, you have to skip

src/kirin/dialects/ilist/rewrite/flatten_add.py
```diff
@@ -0,0 +1,40 @@
...
+            return RewriteResult()
+
+        # rhs:
+        if isinstance((rhs := node.rhs).owner, ilist.New):
+            new_data += rhs.owner.values
+        elif (
+            not isinstance(const_rhs := rhs.hints.get("const"), const.Value)
+            or len(const_rhs.data) > 0
+        ):
+            return RewriteResult()
```
kaihsin, line 36: what if lhs  is ilist.New but rhs is const ilist?

kaihsin: also, if two ilist are different element types, is it ok to do this?

weinbe58: > what if lhs  is ilist.New but rhs is const ilist? 

This is covered in the tests you can merge them here it has to be folded

weinbe58: > also, if two ilist are different element types, is it ok to do this? 

Oh yeah the rewrite should reflect the new elem type

kaihsin (approved):

PR #511, opened 2025-09-28
Removing join.

weinbe58:
Removing join with typeinfer result because it always rewrite statements and so fixpoint will never converge.


Roger-luo:
can we have a test to capture this?

weinbe58:
@Roger-luo I have a test showing that `TypeInfer` never converges. Is that sufficient?

Roger-luo:
I'm convinced this can pass without a test

PR #513, opened 2025-09-29
patch aggressive inline for v0.17

### LINKED ISSUE #508: Aggressive inline fails when kwarg order is not preserved
The following example breaks, where `main_aggressive(1)` returns 1 instead of -1. due to the order of the input kwargs to `foo()` being swapped during aggressive inlining. Removing the top level input argument (`test_aggressive_inline_noargs`) works. This only happens on branch release-0-17; these tests work fine on main.

```

def test_aggressive_inline():
    
    @basic(aggressive=False)
    def foo(arg0, arg1):
        return arg0 - arg1

    @basic(aggressive=True)
    def main_aggressive(arg0):
        return foo(arg1=2, arg0=arg0)
    
    
    main_aggressive.print()
    print(main_aggressive(1))
    assert main_aggressive(1) == -1  

def test_aggressive_inline_noargs():
    
    @basic(aggressive=False)
    def foo(arg0, arg1):
        return arg0 - arg1

    @basic(aggressive=True)
    def main_aggressive():
        return foo(arg1=2, arg0=1)
    
    
    main_aggressive.print()
    print(main_aggressive())
    assert main_aggressive() == -1 
```

neelay893:
Fix #508 (this is fixed in main but needs a patch for release-0-17). Add logic in the `Inline` rewrite rule to permute kwargs.


kaihsin (commented):
how about the test case:

```python

@basic
    def main_aggressive2(param: int):
        def foo3(arg0: int, arg1: int):
            return arg0 - arg1 + param

       alias_foo3 = foo3

        return alias_foo3(arg1=2, arg0=1)


```

neelay893:
@kaihsin added that test case and it passes.

Issue #540, opened 2025-10-23
Propagate source info of statements when inlining


david-pl:
When running the `InlinePass` on a method, the statements that are inlined into the main function body have `stmt.source is None`. It would be nice (e.g. for error handling) to propagate that info.

PR #541, opened 2025-10-23
Ensure SourceInfo is preserved in inline pass and offsets are kepts

rafaelha:
This PR ensures that source info is preserved after inlining or cloning in `.similar()`.

Additionally I found that offsets were lost (which are now added in `lowering.py`).

I tested all of these changes on kirin 0.17.30 together with bloqade-circuit.

Blocks https://github.com/QuEraComputing/bloqade-circuit/pull/552

Addresses #540


kaihsin (approved):
LGTM.

Issue #549, opened 2025-10-27
Circular import error when importing `kirin.dialects.py`


neelay893:
```
In [2]: import kirin.dialects.py
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[2], line 1
----> 1 import kirin.dialects.py

File ~/kirin-workspace/kirin/src/kirin/dialects/py/__init__.py:9
      1 """Python dialects module.
      2 
      3 This module contains a set of dialects that represent
   (...)
      6 entire Python language.
      7 """
----> 9 from . import (
     10     cmp as cmp,
     11     len as len,
     12     attr as attr,
     13     base as base,
     14     list as list,
     15     binop as binop,
     16     range as range,
     17     slice as slice,
     18     tuple as tuple,
     19     unary as unary,
     20     assign as assign,
     21     boolop as boolop,
     22     unpack as unpack,
     23     builtin as builtin,
     24     constant as constant,
     25     indexing as indexing,
     26     iterable as iterable,
     27 )
     28 from .len import Len as Len
     29 from .attr import GetAttr as GetAttr

File ~/kirin-workspace/kirin/src/kirin/dialects/py/len.py:14
     12 from kirin import ir, types, interp, lowering
     13 from kirin.decl import info, statement
---> 14 from kirin.analysis import const
     16 dialect = ir.Dialect("py.len")
     19 @statement(dialect=dialect)
     20 class Len(ir.Statement):

File ~/kirin-workspace/kirin/src/kirin/analysis/__init__.py:23
     17 from kirin.analysis.cfg import CFG as CFG
     18 from kirin.analysis.forward import (
     19     Forward as Forward,
     20     ForwardExtra as ForwardExtra,
     21     ForwardFrame as ForwardFrame,
     22 )
---> 23 from kirin.analysis.callgraph import CallGraph as CallGraph
     24 from kirin.analysis.typeinfer import TypeInference as TypeInference

File ~/kirin-workspace/kirin/src/kirin/analysis/callgraph.py:6
      4 from kirin import ir
      5 from kirin.print import Printable
----> 6 from kirin.dialects import func
      7 from kirin.print.printer import Printer
     10 @dataclass
     11 class CallGraph(Printable):

File ~/kirin-workspace/kirin/src/kirin/dialects/func/__init__.py:21
      9 from kirin.dialects.func.stmts import (
     10     Call as Call,
     11     Invoke as Invoke,
   (...)
[...]

PR #550, opened 2025-10-28
Fix circular import error. issue #549

zhenrongliew:
Fixes bug in issue #549: circular import error when importing `kirin.dialects.py`.

Small edits to `src/kirin/dialects/func/closurefield.py` and `src/kirin/dialects/func/lambdalifting.py`.
Delaying import.

### DIFF
--- a/src/kirin/dialects/func/rewrite/lambdalifting.py
+++ b/src/kirin/dialects/func/rewrite/lambdalifting.py
@@ -1,9 +1,6 @@
-from typing_extensions import TYPE_CHECKING
-
-if TYPE_CHECKING:
-    from kirin.dialects import py
-
 from kirin import ir
+from kirin.passes import TypeInfer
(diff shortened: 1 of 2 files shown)


kaihsin (changes requested):
can you put rewrite under `dialects/func/rewrite`, similar to `dialects/ilist` ?

zhenrongliew:
> can you put rewrite under `dialects/func/rewrite`, similar to `dialects/ilist` ?

Moved. Updated init.py and tests.

src/kirin/dialects/func/rewrite/lambdalifting.py
kaihsin: after chat with @zhenrongliew We don't need to worry about circular import of py and TypeInfer here.

kaihsin (approved):

PR #557, opened 2025-10-30
Ensure inline_getitem returns has_done_something=True only if it has done something

rafaelha:
If `node.result.uses = []` then `node.result.replace_by` actually does not do anything. In such cases, we must return `has_done_something=False` (to avoid infinite fixpoint loops).

This PR also adds some more unit tests.

### DIFF
--- a/src/kirin/emit/abc.py
+++ b/src/kirin/emit/abc.py
@@ -2,30 +2,99 @@ from __future__ import annotations
 
 from abc import ABC, abstractmethod
 from typing import TypeVar
-from dataclasses import dataclass
+from contextlib import contextmanager
+from dataclasses import field, dataclass
 
 from kirin import ir
 from kirin.interp import Frame, abc
+from kirin.idtable import IdTable
+from kirin.worklist import WorkList
 
 TargetType = TypeVar("TargetType")
 
 
 @dataclass
 class EmitFrame(Frame[TargetType]):
-    pass
+    ssa: IdTable[ir.SSAValue] = field(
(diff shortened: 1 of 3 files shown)


src/kirin/dialects/ilist/rewrite/inline_getitem.py
```diff
@@ -25,16 +26,18 @@ def rewrite_Statement(self, node: ir.Statement) -> abc.RewriteResult:
...
+
         index = index_const.data
         if isinstance(index, int) and (
             0 <= index < len(stmt.args) or -len(stmt.args) <= index < 0
         ):
             node.result.replace_by(stmt.args[index])
             return abc.RewriteResult(has_done_something=True)
-        elif isinstance(index, slice):
-            start, stop, step = index.indices(len(stmt.args))
+        elif isinstance(index, (slice, SliceAttribute)):
```
rafaelha: I find that, depending on whether I use, e.g., `my_list[slice(2,4)]` or `my_list[2:4]` I get `slice` or `SliceAttribute`, so I figured I should check for both. Is that expected?

weinbe58: That is quite weird it should lower to the same thing.

rafaelha:
This is blocked by#562 (only this PR should be backported, not #562)

weinbe58 (commented):
It looks fine but I am wondering why this is needed in the first place? In principle fix point rewrites should always run Dead Code elimination to remove these statements that have no uses.

PR #563, opened 2025-11-03
Try to guess element type from values in ilist.New

david-pl:
This was removed as part of #344, so I'm not sure if it should be added back. However, without this change we always get `IList[AnyType]` unless explicitly specifying the element type in `ilist.New`. This breaks a bunch of tests (and code, really) in bloqade-circuit. Adding this back seemed like the much easier change than explicitly setting the `elem_type` everywhere.

### DIFF
--- a/src/kirin/dialects/ilist/stmts.py
+++ b/src/kirin/dialects/ilist/stmts.py
@@ -39,6 +39,9 @@ class New(ir.Statement):
                 elem_type = types.Any
             else:
                 elem_type = values[0].type
+                for v in values[1:]:
+                    elem_type = elem_type.join(v.type)
+
         result_type = IListType[elem_type, types.Literal(len(values))]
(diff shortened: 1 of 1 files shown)


src/kirin/dialects/ilist/stmts.py
```diff
@@ -35,7 +35,10 @@ def __init__(
         elem_type: types.TypeAttribute | None = None,
     ) -> None:
         if not elem_type:
-            elem_type = types.Any
+            if not values:
+                elem_type = types.Any
+            else:
+                elem_type = values[0].type
```
Roger-luo: this is not correct? the type should be union of all elements being used to construct the list.

david-pl: Oh, my bad. Thanks for fixing it!

PR #567, opened 2025-11-05
Remove invalid uses when using ElType in loop typeinfer

### LINKED ISSUE #564: Leftover range iterable after UnrollScf pass
MRE:

```python
from kirin.prelude import structural_no_opt
@structural_no_opt
def main():
    ls = [1,2,3]
    for i in range(1):
        ls[i] = 10
    return ls

UnrollScf(main.dialects).fixpoint(main)
main.print()
```

The resulting IR is

```python
func.func @main() -> Literal(IList([1, 2, 3]),IList[int, Literal(data=3, type=int)]) {
  ^0(%main_self):
  │ %ls = py.constant.constant IList([1, 2, 3]) : !py.IList[!py.int, Literal(3,int)]
  │  %0 = py.constant.constant IList(range(0, 1)) : !py.IList[!py.int, Literal(1,int)]
  │  %i = py.constant.constant 0 : !py.int
  │  %1 = py.constant.constant 10 : !py.int
  │       py.assign.setitem(%ls, %1, %i)
  │       func.return %ls
} // func.func main
```

where the SSA value `% 0 ` is seemingly unused, but it's not eliminated.

This causes rewrites that want to unroll for loops and eliminate all `py.constant` statements (e.g. because the resulting dialect doesn't support it) to fail.

This change was somehow introduced by #382

david-pl:
Closes #564 .

### DIFF
--- a/test/passes/test_unroll_scf.py
+++ b/test/passes/test_unroll_scf.py
@@ -1,5 +1,5 @@
-from kirin.prelude import structural
-from kirin.dialects import py, func
+from kirin.prelude import structural, structural_no_opt
+from kirin.dialects import py, func, ilist
 from kirin.passes.aggressive import UnrollScf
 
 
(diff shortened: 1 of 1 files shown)


src/kirin/dialects/scf/typeinfer.py
```diff
@@ -34,7 +34,10 @@ def for_loop(
         body_block = stmt.body.blocks[0]
         block_args = body_block.args
 
-        eltype = interp_.frame_eval(frame, ElType(stmt.iterable))
+        eltype_stmt = ElType(stmt.iterable)
```
kaihsin: actually.... why are we doing this at first place? can't we just directly infer that from the stmt.iterable lattice element?

kaihsin (approved):

PR #569, opened 2025-11-05
Fix SSAValue usage bug when removing dead blocks

### LINKED ISSUE #565: `CFGCompactify` pass sometimes fails for dead subgraphs
The `DeadBlock` pass, which removes unreachable blocks from the CFG, will sometimes fail when removing dead subgraphs of the CFG. This can happen when a dead block uses an SSAValue from another dead block, and the block with the definition is removed before the block with the use. 

MWE:

```
from kirin.dialects import py

@basic
def deadblock_mwe():
    j = 0
    if False:
        j = 1
        
        if True:
            j = j + 1

        else:
            j = j - 1
    
    return j

deadblock_mwe.print()
deadblock_mwe()
```

results in:
```
/path/to/kirin/kirin/ir/ssa.py(90)delete()
     89             import ipdb; ipdb.set_trace()
---> 90             raise ValueError("Cannot delete SSA value with uses")
     91         self.replace_by(DeletedSSAValue(self))
```

Tested with kirin 0.17.34.

neelay893:
Fix #565 by setting `safe=False` when deleting dead blocks. This prevents checking for usages before converting `SSAValue` to `DeadSSAValue`.

### DIFF
--- a/src/kirin/rewrite/compactify.py
+++ b/src/kirin/rewrite/compactify.py
@@ -27,7 +27,7 @@ class DeadBlock(RewriteRule):
                     self.cfg.predecessors[successor].discard(block)
                 self.cfg.successors.pop(block, None)
                 self.cfg.predecessors.pop(block, None)
-                block.delete()
+                block.delete(safe=False)
                 has_done_something = True
         return RewriteResult(has_done_something=has_done_something)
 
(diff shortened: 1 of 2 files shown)


kaihsin (approved):
LGTM

PR #573, opened 2025-11-07
register vmath desugar as post inference and remove pass

neelay893:
- vmath desugar rewrite (#568) is now registered as a vmath dialect post-inference rewrite rule
- remove vmath/passes.py as it is no longer needed

These changes are separate from (#568) since they are not compatible with kirin 0.17 (in case we want to backport).


kaihsin (approved):
LGTM

Issue #566, opened 2025-11-05
Introduce a ValidationErrorGroup and according stacktrace printing


david-pl:
While looking into validation in bloqade-circuit, we encountered the need to conveniently display multiple error messages at once. A native python `ExceptionGroup` is not sufficient, since it doesn't properly display kirin kernel errors.

Therefore, we introduced a `ValidationErrorGroup` here: https://github.com/QuEraComputing/bloqade-circuit/blob/16429806687591988ea25ee96d2c7c7d35fe9183/src/bloqade/validation/kernel_validation.py#L10.

While the implementation works, and forwards to kirin's exception handler, it still overrides kirin's handler. Therefore, this should be cleaned up and moved to kirin.

PR #581, opened 2025-11-19
Fix ValidationResult to raise all errors

zhenrongliew:
Updated the `raise_if_invalid` method to raise an `ExceptionGroup` containing all validation errors, with a formatted message for better clarity.
Addressing [this issue](https://github.com/QuEraComputing/kirin/issues/566).

### DIFF
--- a/src/kirin/validation/validationpass.py
+++ b/src/kirin/validation/validationpass.py
@@ -54,13 +54,15 @@ class ValidationSuite:
     Caches analysis results to avoid redundant computation when multiple
     validation passes depend on the same underlying analysis.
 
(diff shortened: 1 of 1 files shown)


david-pl:
@zhenrongliew but does this still pretty-print the errors? When I was working on this, using python's native `ExceptionGroup` didn't nicely print the `ir.ValidationError` output, so it's not easy to see the line numbers etc. That's why I went for a dedicated class along with an error handler here: https://github.com/QuEraComputing/bloqade-circuit/blob/d29a1c31f7744f2075dd76732a808e780fba8e51/src/bloqade/validation/kernel_validation.py#L10

Issue #579, opened 2025-11-14
TypeInference bug in `ilist.map`


weinbe58:
Something is going wrong with type inference on the following example:

```python
from kirin.prelude import structural
from kirin.dialects import py, ilist


@structural(typeinfer=True, fold=False, no_raise=False)
def main(n):
    def map_func(i):
        return n + 1
    
    return ilist.map(map_func, ilist.range(4))

main.print()
```

type of the `collection` in ilist.map is bottom. Seems to be somewhat related to type inference of `func.Lambda` but I am not sure.

kaihsin:
fixed in #582

PR #582, opened 2025-11-19
Make MethodType explicit

kaihsin:
This PR address #579 and contain partly of #580


src/kirin/dialects/func/attrs.py
```diff
@@ -12,10 +12,10 @@
 
 from ._dialect import dialect
 
-TypeofMethodType = types.PyClass[Method]
-MethodType = types.Generic(
-    Method, types.TypeVar("Params", types.Tuple), types.TypeVar("Ret")
-)
+# TypeofMethodType = types.PyClass[Method]
```
Roger-luo: Can we get rid of the `Signature` attribute too? Now you have a function type we can just use that to replace the `Signature` attribute

kaihsin:
> I think should be good in general, need to fix a few linting bugs I had in #580

fixed type hint issue

PR #585, opened 2025-11-21
check for Bottom in vmath rewrite

neelay893:
Type inference bugs cause analysis results to `Bottom` in certain cases. Add a check for bottom to not implement vmath binop desugaring if either argument is type `Bottom`.


kaihsin (approved):
LGTM, for context @Roger-luo this is causing some issue with downstream behavior. We should address this typeinfer issue properly in another PR

cduck:
Yay!  This fixed a tricky bug in my pipeline.  Any chance of backporting it to 0.21 @neelay893?

kaihsin:
Yes this should be able to backport to 0.21 too.

PR #632, opened 2026-04-03
ConstProp early termination optimization

johnzl-777:
A (looser) form of this optimization exists in https://github.com/QuEraComputing/bloqade-circuit/pull/736. The idea is to stop constprop from iterating through a loop if `loop_vars` has converged, preventing a bunch of unnecessary future iterations. 

Claude pointed out to me that if there's some dependency on the iteration variable triggering an impure operation in the loop body, the original fix without the guard you currently see would improperly denote the loop as entirely pure. 

I already check for no iteration variable dependency in the bloqade-circuit stuff I'm working on but I don't see the harm in having it duplicated here. 

@Roger-luo agreed this optimization is worth including in Kirin, with kudos to @weinbe58 and @jasonhan3 for pointing this optimization out to me. 

Some benchmarks can be seen here, running against a trivial:

```python
@structural_no_opt
def loop_test(x: int) -> int:
	for _i in range(...):
		x = x + 1
	return x
```

From `range(10)` to `range(30_000)`:

bench_constprop_early_termination

### DIFF
--- a/src/kirin/dialects/scf/constprop.py
+++ b/src/kirin/dialects/scf/constprop.py
@@ -108,14 +108,8 @@ class DialectConstProp(interp.MethodTable):
 
         loop_vars = frame.get_values(stmt.initializers)
 
-        # Early termination is only sound when the loop body does not
-        # reference the iteration variable.  When the body ignores it,
-        # each iteration is a pure function of loop_vars alone, so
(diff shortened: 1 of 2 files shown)


weinbe58 (approved):
LGTM

PR #633, opened 2026-04-04
Adds generic log to the math dialect

### LINKED ISSUE #131: Missing `log` in math dialect.

aadi07:
Fixes #131 - however the issue was caused by inspect.signature throwing a value error when the function passed in has variable arguments (in the case of math.log, the base argument is optional). I've bypassed this by forcing the log to have a base passed in. Not sure this is ideal for the use case here (could be preferable to default to the natural log). Open to any suggestions.


kaihsin:
> When I ran _gen.py under the math dialect, it edited what were formerly references to "lowering" to "lowering2" - it seems to have been an intentional change based on the commit history, but it doesn't seem to be properly recognized, so I also wanted to verify if this was intended or should be kept as lowering?

@aadi07 I think this need to update too. there should be only "lowering" exists now.

aadi07:
Got it, that change should be reflected now as well. Does everything else look good?

src/kirin/dialects/math/__init__.py
```diff
@@ -1,156 +1,123 @@
...
 def expm1(x: float) -> float: ...
 
-
 @lowering.wraps(stmts.fabs)
 def fabs(x: float) -> float: ...
 
-
 @lowering.wraps(stmts.floor)
-def floor(x: float) -> int: ...
+def floor(x: int) -> int: ...
```
kaihsin: This signature is wrong. it should be `floor(x: float) -> int`

src/kirin/dialects/math/__init__.py
```diff
@@ -1,156 +1,123 @@
...
 
-
 @lowering.wraps(stmts.tanh)
 def tanh(x: float) -> float: ...
 
-
 @lowering.wraps(stmts.trunc)
-def trunc(x: float) -> int: ...
-
+def trunc(x: int) -> int: ...
```
kaihsin: same here

src/kirin/dialects/math/interp.py
```diff
@@ -14,176 +13,223 @@ def acos(self, interp, frame: Frame, stmt: stmts.acos):
...
     def floor(self, interp, frame: Frame, stmt: stmts.floor):
         values = frame.get_values(stmt.args)
         return (math.floor(values[0]),)
 
+
+    @impl(stmts.fma)
+    def fma(self, interp, frame: Frame, stmt: stmts.fma):
+        values = frame.get_values(stmt.args)
+        return (math.fma(values[0], values[1], values[2]),)
+
```
kaihsin: please remove fma from this PR. it only support on 3.13+

src/kirin/dialects/math/_gen.py
```diff
@@ -40,12 +40,32 @@ def builtin_math_functions():
...
+                            inspect.Parameter(
+                                "base",
+                                inspect.Parameter.POSITIONAL_OR_KEYWORD,
+                                default=None,
+                                annotation=object,
+                            ),
+                        ],
+                        return_annotation=object,
+                    )
+                    yield name, obj, sig
```
kaihsin, line 62: is there a reason why this is being handled in `except`?

kaihsin (changes requested):

src/kirin/dialects/math/_gen.py
aadi07: log and hypot throw valueerrors in the line sig = inspect.signature(obj), because they have variable number of arguments. I could handle the case outside of except as well, but it has to be handled differently than the other functions. I've overriden the variable behavior, and forced it to have two inputs - as there doesn't seem to be a standard in interp.py, stmts.py, or __init__.py for a function to have variable arguments, but I could try and see how that could be allowed?

src/kirin/dialects/math/__init__.py
aadi07: This is the case with all of trunc, floor, ceil, and any function of the form "is_____" because of lines 136 to 146 in _gen.py which state:
```
for name, obj, sig in builtin_math_functions():
        if "is" in name:
            ret_type = "bool"
        elif name in {"trunc", "ceil", "floor"}:
            ret_type = "int"
        else:
            ret_type = "float"
        f.write(textwrap.dedent(f"""
        @lowering.wraps(stmts.{name})
        def {name}({", ".join(f"{arg}: {ret_type}" for arg in sig.parameters.keys())}) -> {ret_type}: ...
        """))
```
This makes the assumption that the return type and argument types of every built in math function are the same, which is not the case. I'll modify this so that you don't have to make manual changes every time you run _gen.py (I'll just encode the manual exceptions directly into this block).

kaihsin: that would be awesome. Thx

aadi07: I also noticed that the list of exempted functions at the top of _gen.py, aside from the ones that are exempted due to a version issue, seem to exempted due to having integer inputs. If the only reason these weren't include is an inconvenience in generalizing their signatures, I could include them now?
```
# skip some special cases for now
        if name in (
            "prod",
            "perm",
            "modf",
            "ldexp",
            "lcm",
            "isqrt",
            "isclose",
            "gcd",
            "fsum",
            "frexp",
            "factorial",
            "acosh",
            "comb",
            "dist",
            "sumprod",
            "nextafter",
            # 3.10 compat
            "cbrt",
            "exp2",
            # 3.13 compat
            "fma",
        ):
```

kaihsin:
I think some of them are because of version. Some of them are just can be
implemented with subroutine (for example factorial). If you want to keep it
clean in this PR, let's add fma to that list at least

aadi07:
I think the latest commit should have resolved this. I've added fma to the excepted functions list, and made sure that all input parameters are encoded as floats by default, with only the outputs varying appropriately.

kaihsin (approved):

kaihsin:
@aadi07 can you fix the linting issue? you can use pre-commit tool to do that for you

aadi07:
> @aadi07 can you fix the linting issue? you can use pre-commit tool to do that for you

The latest should have addressed the linting issue as well.

aadi07:
Apologies, missed a file while linting, which should now be fixed as well.

aadi07:
The two workflow checks that are failing seem to be unrelated to the code I changed unless I am mistaken? Is there something modification that I'm missing?

Roger-luo:
Thanks for the contribution! I just fixed the CI.

PR #635, opened 2026-04-09
fix `Region.clone()` to use reverse-post-order traversal

neelay893:
Aggressive fold over multiple iterations (fixpoint), can sometimes produce a region with an out of order block sequence (even if the CFG is still well formed). This can cause issues when using `Region.clone()`, where a block using some SSA value is cloned before the block defining it, resulting in the use SSA still pointing to the old object. 

This PR ~~changes `Region.clone()` to traverse blocks in reverse-post-order to avoid this~~ adds a `SortBlocks` rewrite rule to sort the blocks in each region in reverse-post-order. `SortBlocks` is applied at the end of the `CompactifyRegion` pass.

### DIFF
--- a/src/kirin/ir/nodes/region.py
+++ b/src/kirin/ir/nodes/region.py
@@ -20,6 +20,47 @@ if TYPE_CHECKING:
     from kirin.serialization.core.serializationunit import SerializationUnit
 
 
+def _rpo_block_order(region: Region) -> list[Block]:
+    """Return blocks in reverse post-order of the CFG.
+
+    RPO guarantees that in well-formed SSA, a block's dominator is visited
+    before the block itself, so statement results are cloned before their uses.
+    Unreachable blocks (not in CFG) are appended at the end in original order.
+    """
+    from kirin.analysis.cfg import CFG
+
+    cfg = CFG(region)
+    successors = cfg.successors
+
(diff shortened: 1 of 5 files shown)


Roger-luo (changes requested):
I think this is a good change to canonicalize the order of iteration, but I believe it should not be part of the IR iterator but should be a rewrite pass - otherwise we need to pay the price everytime we try to iterate the blocks

Roger-luo:
@neelay893  can you share your claude prompts in the PR for future reference? the PR itself looks good.

Roger-luo (approved):

neelay893:
Claude prompts:

❯ there is an issue w/ Region.clone() where the blocks are traversed in index order, not topologically sorted CFG order. This is usually fine, but can cause problems when fixpointing aggressive fold. See kirin_aggressive_fold_mwe.py for an example, and kirin_region_clone_bug.md for a longer description.
[I'd previously asked codex to debug and issue and generate an MWE and markdown file describing the bug]

[it then made a plan which we went back and forth on for a bit; I can try to find the full chat history if this is of interest]

❯ please review the changes

❯ please refactor the RPO traversal into a rewrite pass that mutates the block sequence within the region, such that subsequent traversals will be in the correct order

Issue #656, opened 2026-05-22
Issue for ir.Method in IList  lowering


kaihsin:
The following mwe produce the error when lowering from python syntax/AST

MWE: 
```python 
from kirin.prelude import basic

@basic
def main():
    return 3

@basic
def main2():    
    return 4


@basic
def main_final():
    return [main, main2]

```

Error msg:
```
Traceback (most recent call last):
  File "/home/kaihsinwu/kirin-workspace/tst.py", line 13, in <module>
    @basic
     ^^^^^
  File "/home/kaihsinwu/kirin-workspace/kirin/src/kirin/ir/group.py", line 274, in __call__
    return wrapper(py_func)
           ^^^^^^^^^^^^^^^^
  File "/home/kaihsinwu/kirin-workspace/kirin/src/kirin/ir/group.py", line 228, in wrapper
    code = self.lowering.python_function(py_func, lineno_offset=lineno_offset)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/kaihsinwu/kirin-workspace/kirin/src/kirin/lowering/python/lowering.py", line 72, in python_function
    region = self.run(
             ^^^^^^^^^
  File "/home/kaihsinwu/kirin-workspace/kirin/src/kirin/lowering/python/lowering.py", line 112, in run
    self.visit(state, stmt)
  File "/home/kaihsinwu/kirin-workspace/kirin/src/kirin/lowering/python/lowering.py", line 147, in visit
    return self.registry.ast_table[name].lower(state, node)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/kaihsinwu/kirin-workspace/kirin/src/kirin/lowering/python/dialect.py", line 72, in lower
    return getattr(self, f"lower_{node.__class__.__name__}", self.unreachable)(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/kaihsinwu/kirin-workspace/kirin/src/kirin/dialects/lowering/func.py", line 63, in lower_FunctionDef
    func_frame.exhaust()
  File "/home/kaihsinwu/kirin-workspace/kirin/src/kirin/lowering/frame.py", line 159, in exhaust
[...]

PR #657, opened 2026-05-22
fix(types.py): Addint is_subseteq_TypeVar to FunctionType to make sure it will be compabitle with generics

weinbe58:
Related issue #656 The problem there is when the FunctionType interaces with the IList Generic type it exposed that we're missing a visitor for `is_subseteq_TypeVar` because during lowering we haven't inferred the type yet.


kaihsin (approved):

kaihsin:
cc @paige-quera @cduck

PR #677, opened 2026-06-23
Interpreter associated types

zhenrongliew:
## Summary

This PR focuses on interpreter refactor to consolidate type parameters :

- collapses `Interpretable<L, I, F, C, E, T>` into `Interpretable<I>`
- moves `Value`, `Error`, and statement `Effect` onto `Interp` associated types
- keeps `Ctx` as the statement-local read/write helper
- makes `ForwardEffect` the forward-only statement-to-frame algebra
- removes the framework-owned `Scope` / `ScopeHook` abstraction
- restores structured traversal as dialect-owned frames: `scf.if` and `scf.for` now push SCF-owned concrete/abstract frames

## Notes

`Enter` / `EnterAny` are gone. Their abstract behavior is now owned by SCF frames. New focused constprop tests cover unknown `scf.if` joining:

- same constants join to `Const(1)`
- different constants join to `Top`

Backward/liveness analysis is intentionally deferred to a follow-up PR with its own effect algebra.


crates/kirin-arith/src/interpreter.rs
```diff
@@ -1,89 +1,72 @@
...
+    I::Error: From<DivisionByZero>,
     T: CompileTimeValue,
 {
-    fn interpret(
-        &self,
-        _location: Location,
-        env: kirin_interpreter::EnvIndex,
-        interp: &mut I,
-    ) -> Result<StatementEffect<F, C, X>, E> {
+    fn interpret(&self, ctx: &mut Ctx<'_, I>) -> Result<I::Effect, I::Error> {
```
Roger-luo: do we plan to merge `I` with `Ctx` in a separate PR? IIUC `I` in practice is a dummy placeholder type for specialization right now?

zhenrongliew: Yes, i think merging`I` with `Ctx` is right.
I tested the current code, and it does not compile:

```rust
impl<I, T> Interpretable<I> for Arith<T>
where
    I: ForwardInterp,
{ ... }

impl<I, T> Interpretable<I> for Arith<T>
where
    I: LiveInterp,
{ ... }
```
Rust reports E0119 conflicting implementations, even if `ForwardInterp` and `LiveInterp` pin different associated `Effect` types, the trait solver does not use that to prove the impls are disjoint.

A merged `I` and `Ctx` redesign could look like:
```rust
trait InterpretCtx {
    type Value;
    type Error;
    type Effect;
}

trait ForwardCtx: InterpretCtx {
    fn read(...);
    fn write(...);
}

trait LivenessCtx: InterpretCtx {
    fn live_after(...);
    fn use_def(...);
}
```
Then forward and liveness can use distinct concrete context types, so Rust can see the impls are disjoint, e.g. conceptually:
```rust
struct ForwardContext<'a, I> { ... }
struct LivenessContext<'a, I> { ... }

impl<'a, I, T> Interpretable<ForwardContext<'a, I>> for Arith<T> { ... }

impl<'a, I, T> Interpretable<LivenessContext<'a, I>> for Arith<T> { ... }
```

zhenrongliew: Did not literally remove the engine type `I`, because it still owns the actual interpreter state: env, frames, linker, value/error/effect types, etc. But moved the dialect-facing specialization from `I` to the concrete context type.

Issue #667, opened 2026-06-12
liveness analysis


Roger-luo:
implement a new liveness analysis using the interpreter framework to test if all the abstraction works well backwards, similar to constprop, we will create this in a separate crate `kirin-liveness`, once we have this implemented, we should be relatively converged on the abstract interpreter framework.

Roger-luo:
we are gonna merge #677 and #682 first there are some remaining small issues to solve

- rename Region -> CFG given now this is kinda specific to CFG
- investigate why interpreter framework is specialized to Region, it should be generic to bodies like `Block`, `DiGraph`, `UnGraph`, a good way to exam this is test it on a toy langauge that combines a computational graph in `DiGraph` with a regular SSA IR (e.g the arith + cf + function region)
- simplify the boilerplate in toy-lang, right now sparse liveness is only useful for validating the IR, if IR is validate the usedef chain is already in the IR attached to each SSAValue, no need to rerun the analysis before dense liveness

PR #683, opened 2026-07-15
Rename Region to CFG

zhenrongliew:
Part of #667. This addresses the first follow-up identified in the issue:

> Rename Region → CFG, given that the current representation is specifically a control-flow graph.

## Summary

- Renames the `Region` terminology to `Cfg` throughout the IR, interpreter, analyses, builders, tests, and documentation.
- Updates downstream consumers to use the new CFG terminology.

This makes it explicit that the representation consists of basic blocks connected by control-flow edges, rather than using the more general term “region.”

## Stack

This is the first PR in the issue #667 follow-up stack. The subsequent PR generalizes interpreter traversal beyond CFG-bodied callables.


Roger-luo:
should it be CFG instead of Cfg?

PR #693, opened 2026-08-05
Fix dataflow analysis worklist algorithm

neelay893:
# Bug

The following kernel:

```python
@basic_no_opt
def augmented_assignment(should_increment: bool) -> int:
    counter = 0
    for _ in range(LOOP_COUNT):
        if should_increment:
            counter += 1
    return counter
```
constant-folds incorrectly to `%counter = 1`. 

This is due to a bug in the way the dataflow worklist algorithm tracks whether a block has been visited; currently this is keyed by `(<block>, <block_args>)`. So, if a block contains SSA dependencies that are not block args, changes in these dependencies will not result in that block being revisited, leading to incorrect analysis results. 

Returning to the above example, the offending lowered IR blocks are:
```
# if statement conditional
^1(%_, %should_increment_1, %counter_1):
    cf.cond_br %should_increment_1 : !py.bool goto ^2(%should_increment_1 : !py.bool) 
        else ^3(%counter_1 : !py.int, %should_increment_1 : !py.boo # go to ^2 if %should_increment_1

# if statement body:
^2(%should_increment_2)
    %counter_2 = !py.add(%counter_1, 1);
```
`^2` does not include `%counter_1` in its block args, so when `%counter_2` widens to `Unknown()` (due to the for loop), `^2` does not get re-queued.

# Fix

This PR adds 1) a `SSAValue --> set[Block]` dependency map to the `SSACfg` abstract interpreter; and 2) a query of this map in the worklist algorithm to re-queue any (reached) dependent blocks after an SSAValue has been modified.


src/kirin/dialects/ssacfg.py
```diff
@@ -73,12 +117,34 @@ def ssacfg(
...
         return result
 
+    @staticmethod
+    def _get_dependents(node: ir.Region) -> dict[ir.SSAValue, set[ir.Block]]:
+        dependents: dict[ir.SSAValue, set[ir.Block]] = {}
+        for stmt in node.walk():
+            block = Abstract._get_enclosing_block(node, stmt)
+            if block is None:
+                continue
+            for value in stmt.args:
```
zhenrongliew: `_get_dependents` walks `stmt.args` (use-def edges) and inverts them to build a def→blocks map — but `SSAValue.uses` is already that def-use index, then just call `enclosing_block` on that.

zhenrongliew (commented):
small fix regarding the def→blocks map.

src/kirin/dialects/ssacfg.py
neelay893: Should be fixed now.

zhenrongliew (commented):
LGTM. Once the GitHub Actions incident is resolved, we'll rerun the jobs.

zhenrongliew (approved):

PR #707, opened 2026-08-18
fix(deserializer): JSON deserialization does not restore SSA reverse-use links

### LINKED ISSUE #706: JSON deserialization does not restore SSA reverse-use links
# JSON deserialization does not restore SSA reverse-use links

## Summary

JSON-deserialized Kirin methods retain their forward operand references, but
the referenced SSA values have empty ``.uses`` sets.  This violates the SSA
use-def invariant and makes standard transformation passes that rely on
reverse-use information unsafe.  In particular, ``Fold`` can run dead-code
elimination and remove values that are still used by later statements,
leaving dangling operands in the IR.

This is independent of the fact that JSON serialization currently omits SSA
``const`` hints: restoring hints with ``HintConst`` works because that pass
does not use or mutate reverse-use bookkeeping.

## Related but separate: SSA constant hints

Result-value ``hints`` (including ``hints["const"]``) are also absent after a
JSON round trip. That can affect analyses that call ``maybe_const`` on a
decoded method, such as measurement-index analysis.

This issue should not require the generic deserializer to serialize or
recompute those hints. Hints are derived analysis metadata rather than
authoritative IR, and the appropriate dialect-aware pass should rebuild them
after decoding:

```python
from kirin.passes import HintConst

HintConst(method.dialects)(method)
```

The reverse-use bug is different: ``.uses`` is structural SSA bookkeeping,
not optional analysis metadata. It must be restored by deserialization before
any client pass runs.

## Minimal reproduction

```python
from kirin import ir
from kirin.passes import Fold
from kirin.prelude import basic_no_opt
from kirin.serialization import JSONSerializer


@basic_no_opt
def kernel(x: int) -> int:
    y = x + 1
    return y * 2


encoded = JSONSerializer().encode(kernel.dialects.encode(kernel))
decoded = basic_no_opt.decode_json(encoded)

# Every operand must be represented by a reverse use on its defining value.
for stmt in decoded.code.walk():
    for index, operand in enumerate(stmt.args):
        assert ir.Use(stmt, index) in operand.uses

# Currently, the assertion above fails. Running Fold can consequently erase
# still-used values through DeadCodeElimination.
Fold(decoded.dialects)(decoded)
```

For a method containing an ``ilist.new`` value consumed by a later statement,
the original method reports one use for that result, while the round-tripped
method reports zero uses even though its printed IR still shows the operand:

```text
original:     %values = py.ilist.new(...); consumer(%values)  # uses = 1
deserialized: %values = py.ilist.new(...); consumer(%values)  # uses = 0
```

``Fold`` includes ``DeadCodeElimination``. It therefore removes ``%values``
as dead, but leaves ``consumer(%values)`` behind. An analysis or interpreter
subsequently fails with an error such as:

```text
InterpreterError: SSAValue <ResultValue ...> not found
```

## Expected behavior

A deserialized ``ir.Method`` should satisfy the same SSA use-def invariants as
the source method. In particular, for every statement operand ``operand`` at
position ``index``:

```python
ir.Use(statement, index) in operand.uses
```

``Fold`` should be safe to run on a successfully decoded method and should
preserve its semantics.

## Likely cause

``BaseDeserializer.deserialize_statement`` directly assigns deserialized
operands to ``out._args``:

```python
out._args = self.deserialize_tuple(data["_args"])
```

That bypasses ``Statement.args``' public setter, which registers
``ir.Use(statement, index)`` on each operand.

## Proposed fix

Reconstruct statement operands through the normal argument-setting path (or
explicitly rebuild all use records) during deserialization. For example, after
initializing the fresh statement's argument tuple:

```python
out._args = ()
out.args = self.deserialize_tuple(data["_args"])
```

The exact implementation may prefer a dedicated deserialization helper, but
it must establish all reverse-use links before returning the decoded method.

## Regression tests

Add a serializer round-trip test that:

1. Verifies every deserialized statement operand has its matching ``ir.Use``
   entry in ``operand.uses``.
2. Runs ``Fold`` on a deserialized method with values used by later
   statements.
3. Verifies the transformed method remains valid and has the same interpreter
   result as the original method.

zhenrongliew:
Deserializer assigned `out._args` directly bypassing the `args` setter that calls `arg.add_use(Use(self, idx))`.

### Fixed by:
```
out._args = ()
out.args = self.deserialize_tuple(data["_args"])
```
`_args` must be seeded first because the setter iterates it to drop stale uses, and the statement comes from `Statement.__new__` with no attributes set.

Closes #706

### DIFF
--- a/src/kirin/serialization/base/deserializer.py
+++ b/src/kirin/serialization/base/deserializer.py
@@ -142,7 +142,8 @@ class Deserializer:
         self._ctx.Statement_Lookup[data["id"]] = out
         out.dialect = self.deserialize(data["dialect"])
         out.name = self.deserialize_str(data["name"])
-        out._args = self.deserialize_tuple(data["_args"])
+        out._args = ()
+        out.args = self.deserialize_tuple(data["_args"])
         out._results = self.deserialize_list(data["_results"])
         out._name_args_slice = self.deserialize_dict(data["_name_args_slice"])
         out.attributes = self.deserialize_dict(data["attributes"])
--- a/test/serialization/test_use_def.py
+++ b/test/serialization/test_use_def.py
@@ -1,0 +1,180 @@
+"""Regression tests for https://github.com/QuEraComputing/kirin/issues/706.
+
+Deserialization used to assign `Statement._args` directly, bypassing the
+`Statement.args` setter that registers `ir.Use(stmt, index)` on each operand.
+The decoded IR therefore had correct forward operand references but empty
+`SSAValue.uses` sets, which makes `DeadCodeElimination` (and hence `Fold`)
+delete values that are still used.
+"""
+
+import pytest
+
+from kirin import ir
+from kirin.passes import Fold, HintConst
+from kirin.prelude import basic, basic_no_opt
+from kirin.dialects import ilist
+from kirin.serialization.bsonserializer import CompressedBSONSerializer
+
+
+@basic_no_opt
+def straight_line(x: int) -> int:
+    y = x + 1
+    return y * 2
+
+
+@basic_no_opt
+def branching(x: int, flag: bool) -> int:
+    if flag:
+        y = x + 1
+    else:
+        y = x - 1
+    return y * 2
+
+
+@basic_no_opt
+def looping(n: int) -> int:
+    acc = 0
+    for i in range(n):
+        acc = acc + i
+    return acc
+
+
+@basic_no_opt
+def ilist_consumer(x: int):
+    values = ilist.IList([x, x + 1, x + 2])
+    return values[0] + len(values)
+
+
+@basic_no_opt
+def nested_closure(y: int):
+    def inner(x: int):
+        return x * y + 1
+
+    return inner(3)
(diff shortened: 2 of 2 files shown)


johnzl-777 (approved):

src/kirin/serialization/base/deserializer.py
```diff
@@ -142,7 +142,8 @@ def deserialize_statement(self, serUnit: SerializationUnit) -> ir.Statement:
         self._ctx.Statement_Lookup[data["id"]] = out
         out.dialect = self.deserialize(data["dialect"])
         out.name = self.deserialize_str(data["name"])
-        out._args = self.deserialize_tuple(data["_args"])
```
kaihsin, line 145: Shouldn't this also need to modify on the serialize side?

zhenrongliew: No change needed on the serialize side. Reading from the private `_args` gives the same data as the public `args`.
Only deserialize needs to write, and has to go through the public `args`.

PR #711, opened 2026-08-26
Reduce serialized IR size and round-trip cost

zhenrongliew:
### Problem
Serialization had redundant metadata. ~76% of every payload is envelope metadata (`__serialization_unit__,` `module_name`, `class_name`; all derivable from kind for core types), SSA values are re-serialized in full at every use site, and repeated type attributes are written out once per mention.
### Changes

1. Compact wire codec.  Units serialize as `{"$u": [kind, data]}`, with `module_name/class_name` elided only when the pair matches a reviewed static table; everything else keeps its full descriptor. `{"$m": [...]} `escapes user mappings that would otherwise resemble a control tag.
2. SSA operand references.  A value's full record is written once at its owner position (`block._args`, `stmt._results`); operands are `ssa_ref` units resolved through the decoder's existing `SSA_Lookup`.
3. `TypeAttribute` identity references. Repeated `TypeAttribute` identities emit one definition plus `attr_ref` units. `PyAttr` and other general attributes are unchanged.

### Rsults

|  | JSON size | gzip'd BSON size | JSON round trip | gzip'd BSON round trip |
|---|---:|---:|---:|---:|
| **mean across test kernels** | **−76.7%** | **−24.1%** | **29.6% faster** | **8.4% faster** |

### DIFF
--- a/src/kirin/serialization/base/context.py
+++ b/src/kirin/serialization/base/context.py
@@ -17,11 +17,20 @@ class MethodSymbolMeta(TypedDict, total=False):
 
 
 @dataclass
+class TypeAttributeSerializationEntry:
+    owner: types.TypeAttribute
+    unit: SerializationUnit
+
+
+@dataclass
 class SerializationContext:
     ssa_idtable: IdTable[ir.SSAValue] = field(default_factory=IdTable[ir.SSAValue])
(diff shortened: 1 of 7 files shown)


kaihsin (approved):

weinbe58 (approved):
LGTM

PR #713, opened 2026-08-28
fix(serializer): preserve TypeAttribute reference identity

zhenrongliew:
## Summary
Serialization should preserve identity relationships:

Original IR | (De)serialized-roundtrip IR
-- | --
Same instance referenced twice | One new instance referenced twice
Two distinct but equal instances | Two distinct but equal new instances

## Motivation

#711 encodes repeated `TypeAttribute` instances as one definition followed by
`attr_ref`s. However, the decoder recreated the referenced definition for every
`attr_ref`, preserving v1's value behavior but losing the identity relationship of the original IR.

As @weinbe58 demonstrated, this can also make decoding exponential:

```python
typ = types.Generic(tuple, typ, types.Vararg(typ))
```
Each level references the previous type twice. The serialized DAG is linear, but
recreating every reference expands it into a binary tree during decoding.

## Fix
Add a `TypeAttribute_Lookup` instance table to `SerializationContext`. Each
`TypeAttribute` definition is decoded once and cached by its serialization ID;
subsequent `attr_refs` return that same instance.

### DIFF
--- a/src/kirin/ir/attrs/types.py
+++ b/src/kirin/ir/attrs/types.py
@@ -368,7 +368,11 @@ class Literal(TypeAttribute, typing.Generic[LiteralType], metaclass=LiteralMeta)
     ) -> "Literal":
         d = deserializer.deserialize(serUnit.data["value"])
         type_attr = deserializer.deserialize(serUnit.data["type"])
-        return Literal(d, type_attr)
+        # Reconstruct the serialized graph without reusing a process-local
+        # LiteralMeta cache entry that may reference a different type instance.
+        literal = object.__new__(cls)
+        literal.__init__(d, type_attr)
+        return literal
 
 
 @typing.final
--- a/src/kirin/serialization/base/context.py
+++ b/src/kirin/serialization/base/context.py
@@ -34,6 +34,7 @@ class SerializationContext:
 
     Dialect_Lookup: dict[str, ir.Dialect] = field(default_factory=dict)
     SSA_Lookup: dict[str, ir.SSAValue] = field(default_factory=dict)
+    TypeAttribute_Lookup: dict[str, types.TypeAttribute] = field(default_factory=dict)
     Statement_Lookup: dict[str, ir.Statement] = field(default_factory=dict)
     Block_Lookup: dict[str, ir.Block] = field(default_factory=dict)
     Region_Lookup: dict[str, ir.Region] = field(default_factory=dict)
@@ -56,6 +57,7 @@ class SerializationContext:
 
     def clear(self) -> None:
         self.SSA_Lookup.clear()
+        self.TypeAttribute_Lookup.clear()
         self.Block_Lookup.clear()
         self.Region_Lookup.clear()
         self.Statement_Lookup.clear()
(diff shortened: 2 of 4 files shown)


weinbe58:
Verified the fix separately — the exponential decode is gone (flat ~0.5 ms across depths 8–22, vs ~43 s at depth 22 before), encode is unaffected, and the full suite is green.

One coverage regression worth fixing before merge: rewriting `test_nested_type_graph_preserves_v1_value_semantics` into `test_nested_type_graph_preserves_identity_relationships` dropped the only `Union`/`Literal` coverage in the serialization tests. The old version built `types.Literal(1, leaf)` and `types.Union(literal, types.String)`; the new one only exercises `Generic`. After this PR, `types.Union` and `types.Literal` appear in no serialization test anywhere under `test/` — on `main` they're at `test_type_attribute_refs.py:190-191`.

Worth keeping because those two have the least ordinary construction paths, and they interact with the new `TypeAttribute_Lookup` caching in ways `Generic` doesn't exercise: `UnionMeta.__call__` does subset-simplification and can return a *member* instead of a `Union`, and `LiteralMeta` interns on `(data, datatype)`. A small separate test that round-trips a nested `Union`/`Literal` graph and asserts the shared leaf comes back shared would cover it — no need to restore the old test's shape.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

zhenrongliew:
> Verified the fix separately — the exponential decode is gone (flat ~0.5 ms across depths 8–22, vs ~43 s at depth 22 before), encode is unaffected, and the full suite is green.
> 
> One coverage regression worth fixing before merge: rewriting `test_nested_type_graph_preserves_v1_value_semantics` into `test_nested_type_graph_preserves_identity_relationships` dropped the only `Union`/`Literal` coverage in the serialization tests. The old version built `types.Literal(1, leaf)` and `types.Union(literal, types.String)`; the new one only exercises `Generic`. After this PR, `types.Union` and `types.Literal` appear in no serialization test anywhere under `test/` — on `main` they're at `test_type_attribute_refs.py:190-191`.
> 
> Worth keeping because those two have the least ordinary construction paths, and they interact with the new `TypeAttribute_Lookup` caching in ways `Generic` doesn't exercise: `UnionMeta.__call__` does subset-simplification and can return a _member_ instead of a `Union`, and `LiteralMeta` interns on `(data, datatype)`. A small separate test that round-trips a nested `Union`/`Literal` graph and asserts the shared leaf comes back shared would cover it — no need to restore the old test's shape.
> 
> Co-Authored-By: Claude Opus 5 [noreply@anthropic.com](mailto:noreply@anthropic.com)

Restored separate Union/Literal structural round-trip coverage and added a test where a TypeVar is shared between a `Literal` and a `Generic`.
The identity test exposed an issue: `Literal.deserialize` called the normal `Literal(...)` constructor, whose `LiteralMeta` cache could return the source `Literal`. That retained the source leaf instead of the decoded leaf, so the objects were structurally equal but the shared instance was lost.
I updated `Literal.deserialize` to reconstruct a fresh instance without consulting the process-local interning cache.

weinbe58 (approved):
LGTM

PR #717, opened 2026-09-01
Optimize shared-method serialization with method references

zhenrongliew:
### Summary
This is a breaking payload-representation change. Methods are now defined using identity-based IDs, repeated call sites use `method_ref`, and the previous `symbol_table`/mangled-method representation is removed.
Payloads containing the old mangled method records are not supported by this decoder.
Explicit codec versioning and the policy for rejecting unversioned payloads are separated into the stacked follow-up PR #720

### Motivation
Previously, every call site serialized and reconstructed the complete method wrapper, body already de-duplicated with `region_ref`.
On the representative shared-callee benchmark, relative to #713:
- JSON: 61,470 → 42,459 bytes (−31%)
- CBSON: 5,021 → 4,566 bytes (−9%)
- JSON round trip: 4.36 → 2.31 ms (−47%)
- CBSON round trip: 6.22 → 4.24 ms (−32%)

### DIFF
--- a/src/kirin/ir/attrs/py.py
+++ b/src/kirin/ir/attrs/py.py
@@ -100,8 +100,8 @@ class PyAttr(Data[T]):
     def deserialize(
         cls: Type["PyAttr"], serUnit: "SerializationUnit", deserializer: "Deserializer"
     ) -> "PyAttr":
-        pytype = deserializer.deserialize(serUnit.data["pytype"])
         value = deserializer.deserialize(serUnit.data["data"])
+        pytype = deserializer.deserialize(serUnit.data["pytype"])
         return PyAttr(value, pytype=pytype)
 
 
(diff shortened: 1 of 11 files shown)


kaihsin (approved):
I like this, its more clean than previously. Notice that the change of how schema version is assigned will be super breaking. @david-pl @kylequera.

kaihsin:
consider this is breaking change, Dennis, can you decouple the versioning change into another PR?

zhenrongliew:
> consider this is breaking change, Dennis, can you decouple the versioning change into another PR?

Split into: #717 now contains only the method-reference payload change. Explicit codec versioning has moved to the stacked PR #720.

Both PRs are breaking changes.
#717 is payload-format breaking because it removes the previous `symbol_table`/mangled-method representation. 
#720 formalizes the new representation as codec v2 and rejects unsupported unversioned payloads before decoding.

kaihsin:
Thanks I will remove the breaking from this PR then

Roger-luo (approved):

PR #684, opened 2026-07-15
Generalize interpreter traversal over CFG, Block, DiGraph, and UnGraph bodies

zhenrongliew:
Generalize interpreter body traversal beyond CFGs

Part of #667. This addresses the second follow-up identified in the issue:

> Investigate why the interpreter framework is specialized to Region/CFG. It should support bodies such as Block, DiGraph, and UnGraph, demonstrated with a language combining computational graphs and regular SSA IR.

Depends on the Region → CFG rename PR.

## Summary

- Introduces a closed `Body` vocabulary covering:
  - `Cfg`
  - `Block`
  - `DiGraph`
  - `UnGraph`
- Allows callable statements to own CFG, Block, DiGraph, or UnGraph bodies.
- Separates body representation traversal from invocation context:
  - `CfgFrame` traverses CFGs.
  - `BlockFrame` traverses one Block.
  - `DiGraphFrame` traverses a directed computational graph.
  - `CallFrame` manages callee resolution, activation lifetime, returns, and caller result binding.
- Adds a default topological DiGraph walker and reports an error for directed cycles.
- Provides no universal UnGraph walker. A compiler supplies UnGraph scheduling policy through `FrameBuild::from_ungraph_entry`; otherwise the interpreter reports `NoDefaultWalker`.
- Adds a mixed test language combining ordinary CFG SSA, callable Blocks, directed computational graphs, structured Blocks, and callable UnGraph policy.

## Scope

`Body` remains an intentionally closed enum of Kirin’s built-in body representations. The extensibility point for UnGraph is its traversal policy, since an undirected graph has no inherent execution order.

The remaining toy-lang sparse/dense liveness simplification from #667 is left for a separate PR.

PR #723, opened 2026-09-03
refactor(interpreter): invert frame composition ownership

zhenrongliew:
Stacked on #684.  Make each interpreter engine's private stack enum own conversion from reusable member frames. 

### Problem:
Concrete member frames depended on the total enum that stores them. BlockFrame required `F: FrameBuild<V, E>` and re-wrapped itself with `F::from_block(self)`, so the dependency ran:

`BlockFrame -> FrameBuild -> total enum -> BlockFrame`
The cause was in the frame protocol itself. `Frame<I, F>`'s three methods all returned `FrameEffect<F, C>` — over `F`, never over `Self`. A reusable frame was therefore obliged to construct the enclosing enum's variant for itself, which it could only do through a construction trait the enum had to implement. That forced a chain of scaffolding: a `FrameBuild` derive per engine family, a `Build*` construction trait per dialect frame.

### Fix:
Split the parent from the child in `FrameEffect`:

```rust
pub enum FrameEffect<P, C, F = P> {
    Continue(P),                    // P = this frame's own next state
    Push { parent: P, child: F },   // F = configured representation of a child
    Done,
    Complete(C),
}

pub trait Frame<I: FrameEngine, F = Self> {
    fn step_into(self, interp: &mut I) -> Result<FrameEffect<Self, Self::Completion, F>, I::Error>;
    // ...
}
```
A member continuation now returns itself for `Continue` and as the suspended parent of `Push`. A private closed stack-item enum is the composition root: it owns the `From<Member>` conversions, dispatches exhaustively, and lifts member state into its own variant with the new `FrameEffect::map_next`.


crates/kirin-derive-interpreter/src/frame.rs
planza-quera, line 1: Is there a reason as to why the snapshot tests are no longer needed for this macro?

planza-quera (approved):

example/toy-lang/src/interpreter/tests.rs
planza-quera, line 1: are these tests also not needed any longer?

planza-quera (approved):

crates/kirin-derive-interpreter/src/frame.rs
zhenrongliew: You're right, the deleted snapshot covered the old `FrameBuild`, `AbstractFrameBuild`, and `DenseFrameBuild`. I've not added new snapshots for the new `frame.rs`.

example/toy-lang/src/interpreter/tests.rs
zhenrongliew: `TracingFrame` wrapped `ToyFrame`. Previously, `ToyFrame` was generic over an outer frame type `F`, and its members returned themselves wrapped in that type through `FrameBuild`/`BuildScf*`.

This PR moves that wrapping responsibility to the composition enum: members return `Self`. I removed the two tests because they were testing the previous pattern.

crates/kirin-derive-interpreter/src/frame.rs
zhenrongliew: Added a snapshot covering member frame conversions, the new `[derive(Frame)]` macro. [6419a99](https://github.com/QuEraComputing/kirin/pull/723/commits/6419a9919fcb7b08968a44aae9c9fa0f9399cddf)

PR #724, opened 2026-09-03
refactor(interpreter): unify callable roots

zhenrongliew:
Stacked on #723.  Unify callable entry across forward and backward interpreters: separate body discovery from forward argument binding, route backward roots through linkers, honor `FunctionTarget.stage`.

### Three problems:
1. Callable-body discovery was entangled with forward argument binding.

`FunctionEntry` was parameterised on the engine and took a value product:

```rust
pub trait FunctionEntry<I: Interp>: Dialect {
    fn function_entry(&self, args: Product<I::Value>, interp: &mut I)
        -> Result<CallableBody<I::Value>, I::Error>;
}
```
This made it inherently incompatible with backward interpreters, the signature permitted argument-sensitive body selection — so a backward interpreters could not discover structure without fabricating a dummy `Product<LiveSet>`.

2. The backward interpreters bypassed the linker
`analyze(stage, body: impl Into<Body>)` skipped symbol/specialization resolution, target-stage selection, and linker policy entirely. 

Both backward engines now take `Lk = SameStageLinker`, hold a linker field, and expose `with_linker`, matching the concrete and forward engines:

```rust
pub fn analyze(&mut self, stage: CompileStage, callee: Callee) -> Result<BodyScope, E>
```
Every engine root now shares one validated prefix, through a single `resolve_callable`:

```
caller stage + Callee
    -> Linker::resolve
    -> FunctionTarget
    -> FunctionEntry dispatch at FunctionTarget.stage
    -> CallableBody
    -> engine-specific boundary initialization
```

3. Dense Liveness graph analysis silently did nothing
`DirectBodyBlocks` in returned `Vec::new()` for DiGraph/UnGraph, so the solver ran to completion over an empty owner set and reported success having done no work.
Tests now assert that both demand and classic liveness reject graph bodies explicitly.


crates/kirin-derive-interpreter/src/function_entry.rs
```diff
@@ -29,26 +29,19 @@ pub fn do_derive_function_entry(input: &syn::DeriveInput) -> darling::Result<Tok
 fn emit_function_entry(
     ctx: &DeriveContext<'_, InterpreterLayout>,
     interp_crate: &syn::Path,
-    ir_crate: &syn::Path,
+    _ir_crate: &syn::Path,
```
planza-quera, line 32: why not remove this argument altogether?

crates/kirin-interpreter/src/core/effect.rs
```diff
@@ -132,35 +132,28 @@ impl From<Symbol> for Callee {
...
+/// `UnGraph` traversal is a dialect/compiler-supplied call-body traversal,
+/// rejected with
 /// [`InterpreterError::NoDefaultWalker`](crate::InterpreterError) when no
 /// policy is provided.
-pub struct CallableBody<V> {
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub struct CallableBody {
     pub body: Body,
-    pub args: Product<V>,
 }
```
planza-quera, line 153: Is `CallableBody` still needed? Could we just use `Body`?

crates/kirin-interpreter/src/engines/sparse_backward/interp.rs
```diff
@@ -603,23 +630,33 @@ where
...
     ///
     /// **Prepass**: walk the body's containment hierarchy, running every
     /// statement's rule once with nothing demanded — impure statements and
     /// terminators contribute the demand roots.
     /// **Propagation**: drain the value worklist; each risen value dispatches
     /// the rules that translate its demand.
-    pub fn analyze(&mut self, stage: CompileStage, body: impl Into<Body>) -> Result<(), E> {
-        let body = body.into();
-        let scope = (stage, body);
+    pub fn analyze(&mut self, stage: CompileStage, callee: Callee) -> Result<BodyScope, E> {
```
planza-quera, line 648: By restricting `analyze` to have a `Callee` instead of `Body` as arg, are we not limiting the kinds of expressions that we can analyze here? e.g. can't analyze a SCF region?

example/toy-lang/src/interpreter/mod.rs
```diff
@@ -224,10 +145,25 @@ pub fn analyze_classic_liveness(
...
+        .lookup_function_by_name(function_name)
+        .ok_or_else(|| InterpreterError::MissingFunctionName(function_name.into()))?;
+    let mut engine: DenseLiveness<
+        '_,
+        Stage,
+        InterpreterError,
         ToyDenseBackwardFrame<LiveSet, InterpreterError>,
-    >(pipeline, stage, cfg)?;
+        CrossStageLinker,
+    > = DenseLiveness::new(pipeline).with_linker(CrossStageLinker);
```
planza-quera, line 160: I might be misunderstanding the kinds of linker, but is this correct to be cross-stage? Shouldn't liveness be done on the stage defined by `stage_name`?

planza-quera (approved):
In the future, please split the commits into the individual "actions". For example, each item in the second commit's description could be its own commit. This will make reviewing a lot easier when the PR is so large.

Given my lack of experience, I could not really review the correctness of the changes too well. My comments may also stem from lack of understanding.

crates/kirin-interpreter/src/core/effect.rs
zhenrongliew: `CallableBody` is removed in the next stacked-PR #729; moved into `kirin-ir` as a structural query. 

Currently the `CallableBody` is designated as the call entry, not just any `Body`.

crates/kirin-derive-interpreter/src/function_entry.rs
zhenrongliew: This file is deleted in the next stacked-PR #729 as well.

crates/kirin-interpreter/src/engines/sparse_backward/interp.rs
zhenrongliew: Nested bodies are still analyzed, just not from the root entry point `analyze()`.

example/toy-lang/src/interpreter/mod.rs
zhenrongliew: Agreed, liveness is now pinned to `stage_name` via `SameStageLinker` (facc20bbc).

PR #729, opened 2026-09-08
refactor(ir): separate callable discovery from interpreter dispatch

zhenrongliew:
Stacked on #724. 

[#724](https://github.com/QuEraComputing/kirin/pull/724)
 made callable-body discovery input value-independent and gave every engine one shared callable-root prefix. It left the discovery mechanism in the interpreter. This PR moves it into `kirin-ir` and makes it derivable from the operation's own declaration.

**Previously:**
Discovering a callable’s body required routing through `InterpDispatch::dispatch_function_entry`.
`FunctionTarget` copied the definition statement, so a linker could contradict the IR. `FunctionTarget { stage, function, definition }` carried a Statement copied out of the specialization record, even though `SpecializedFunctionInfo` authoritatively owns it.
The declaration was hand-written and unchecked. Every language enum added `FunctionEntry` to its derive list and marked `#[callable]` variants; every dialect wrote the `impl` by hand.

**Fix:**
Move callable-body discovery into `kirin-ir`, so interpreters can discover a definition’s implementation without
dispatch. `Dialect` now derives `HasCallableBody`: callable definitions mark one body with `#[kirin(callable_body)]`.

Restrict linking to selecting a stage and specialization (`LinkTarget`).
Structural discovery reads the specialization’s definition at that stage and returns its body alongside the target (`ResolvedCallable`). 

Remove `FunctionEntry`, its derive and wrapper markers, and`InterpDispatch::dispatch_function_entry`. Concrete execution and forward and backward analyses share structural discovery, then apply their own
boundary initialization and traversal.


tests/body_kinds.rs
```diff
@@ -50,7 +50,7 @@ use kirin_test_languages::GraphFunctionLanguage;
 #[derive(Debug)]
 enum TestError {
     Core(InterpreterError),
-    ArithConversion(ArithConversionError),
+    ArithConversion((ArithConversionError)),
```
planza-quera, line 53:
```suggestion
    ArithConversion(ArithConversionError),
```
Accidental change? Seems to be producing a warning

crates/kirin-derive-ir/src/tests/snapshots/kirin_derive_ir__tests__dialect__dialect_derive_custom_crate_path.snap
```diff
@@ -336,3 +336,9 @@ impl kirin_ir::HasSignature<Self> for Nop {
         None
     }
 }
+#[automatically_derived]
+impl kirin_ir::HasCallableBody for Nop {
+    fn callable_body(&self) -> ::core::option::Option<kirin_ir::Body> {
+        ::core::option::Option::None
+    }
+}
```
planza-quera, line 344: I noticed that none of the snapshots check for the case where `#[kirin(callable_body)]` is explicitly declared. Perhaps we want to include another test that checks this is generated properly.

crates/kirin-interpreter/src/core/linker.rs
```diff
@@ -1,65 +1,73 @@
...
+}
+
+/// A linked target together with its discovered Kirin implementation.
+///
+/// Target identity is used for analysis contexts; the body selects the IR to
+/// traverse. Engine-specific boundary initialization follows discovery.
+#[derive(Clone, Copy, Debug)]
+pub struct ResolvedCallable {
+    pub target: LinkTarget,
+    pub body: Body,
```
planza-quera, line 24: Is this body different to the `SpecializedFunction`'s body inside `LinkTarget`?

planza-quera (approved):
Looks like a very nice refactoring. Like in the previous PR I would have appreciated the commit being broken down into smaller units (though I realize you completed this work before I had reviewed it).

tests/body_kinds.rs
zhenrongliew: Yes, accidental change

crates/kirin-interpreter/src/core/linker.rs
zhenrongliew: No, they're the same body. This was intended to save some unnecessary queries through `SpecializedFunction`.

planza-quera: would it make sense to instead implement a getter method for the body and then remove the need for `ResolvedCallable`?

crates/kirin-derive-ir/src/tests/snapshots/kirin_derive_ir__tests__dialect__dialect_derive_custom_crate_path.snap
zhenrongliew: Added two snapshot tests: 
1. A statement with two `Body`-like fields; only one is marked callable. 
2. A language with mixed callable variants.

 ([7a5ddd6](https://github.com/QuEraComputing/kirin/pull/729/commits/7a5ddd6db8a9b8cc2251ac62a2495d0a47f497e4))

crates/kirin-interpreter/src/core/linker.rs
zhenrongliew: You're right, made changes in [fd5d440](https://github.com/QuEraComputing/kirin/pull/729/commits/fd5d44059d96d064a20a3d4e3055471013e85ca9)

Removed `ResolvedCallable`. Added `LinkTarget::body(pipeline)`, which reads the specialization's definition in `target.stage`. Callers keep the target for stage and analysis identity.

Split `CallServices::resolve_callable` into two methods:
`resolve_callee` and `discover_body`.

planza-quera: looks good, thanks!

PR #736, opened 2026-09-17
Fix type inference fallback for statements with several variadic argument groups

nathanieltornow:
`TypeInference.eval_fallback` zips the declared argument fields with the flat argument list. If a variadic group holds more than one value, later arguments meet the next field's type and become `Bottom`. This PR reads each field's own values, so every argument meets the type of its own field.

**Minimal example**

```python
from kirin import ir, types
from kirin.decl import info, statement
from kirin.dialects import func, py
from kirin.passes import TypeInfer
from kirin.prelude import basic_no_opt

dialect = ir.Dialect("demo")


@statement(dialect=dialect)
class Apply(ir.Statement):
    targets: tuple[ir.SSAValue, ...] = info.argument(types.Int)
    params: tuple[ir.SSAValue, ...] = info.argument(types.Float)


group = basic_no_opt.add(dialect)
block = ir.Block()
block.args.append_from(types.Any, "self")
a, b, angle = py.Constant(1), py.Constant(2), py.Constant(0.5)
apply = Apply(targets=(a.result, b.result), params=(angle.result,))
none = func.ConstantNone()
for stmt in (a, b, angle, apply, none, func.Return(none.result)):
    block.stmts.append(stmt)
code = func.Function(
    sym_name="main",
    signature=func.Signature(inputs=(), output=types.NoneType),
    body=ir.Region(block),
)
TypeInfer(group)(ir.Method(dialects=group, code=code, sym_name="main", arg_names=[]))

print(b.result.type)  # before: BottomType(), after: int
```

E.g., Bloqade's stim `NewPauliString` hits this bug.


test/analysis/dataflow/typeinfer/test_variadic_groups.py
```diff
@@ -0,0 +1,51 @@
...
+dialect = ir.Dialect("demo")
+
+
+@statement(dialect=dialect, init=False)
+class Apply(ir.Statement):
+    """Two variadic argument groups of different types."""
+
+    name = "apply"
+    targets: tuple[ir.SSAValue, ...] = info.argument(types.Int)
+    params: tuple[ir.SSAValue, ...] = info.argument(types.Float)
```
zhenrongliew, line 17: `Apply` is group-only, so `f.group` is always `True` and the scalar arm of
`getattr(node, name) if f.group else (getattr(node, name),)` is not tested.

Please add a a scalar field to test that.

test/analysis/dataflow/typeinfer/test_variadic_groups.py
```diff
@@ -0,0 +1,51 @@
+from kirin import ir, types
+from kirin.decl import info, statement
+from kirin.passes import TypeInfer
+from kirin.prelude import basic_no_opt
+from kirin.dialects import py, func
+
+dialect = ir.Dialect("demo")
+
+
+@statement(dialect=dialect, init=False)
```
zhenrongliew: the test could be a bit shorter.

`init=False` plus the hand-written `__init__` can be dropped; and the hand-written `args_slice` isn't quite the one codegen produces
(`{'params': slice(2, None)}` vs `{'params': slice(2, 3)}`)

zhenrongliew (changes requested):
The fix looks good to me.
Just some small edits in the test file.

test/analysis/dataflow/typeinfer/test_variadic_groups.py
nathanieltornow: added a bool in between

test/analysis/dataflow/typeinfer/test_variadic_groups.py
nathanieltornow: removed init 👍

nathanieltornow:
Thanks @zhenrongliew, I've addressed your comments 👍

zhenrongliew (approved):

